PR #459, opened 2026-05-19
agent-1: validate GC membership index values

samuelstroschein:
Summary:
- Validate by_change_membership index values during changelog GC scans.
- Switch the GC membership scan from key-only to full-value projection; values are defined as empty, so corrupt non-empty bytes now fail planning.
- Add a regression that corrupts a membership value, expects collect_garbage to error, and verifies no sweep writes are staged.

Verification:
- cargo test -p lix_engine changelog::gc::tests::gc_errors_when_by_change_membership_value_is_corrupt --quiet
- cargo test -p lix_engine changelog::gc::tests --quiet
- git diff --check

Review note:
- Attempted the required three-agent review loop, but all spawned reviewers failed due to the account usage limit. This draft keeps the stack moving while making that review gap explicit.


samuelstroschein:

PR #460, opened 2026-05-19
agent-1: avoid by-file count heuristic overflow

samuelstroschein:
Summary:
- Avoid overflow in the tracked-state by-file scan fallback heuristic by comparing with division instead of multiplying the corruptible index count.
- Add a regression with a corrupt by-file internal summary reporting u64::MAX rows; the scan falls back to the primary root without touching the bogus child.

Verification:
- cargo test -p lix_engine tracked_state::context::tests::by_file_scan_fallback_heuristic_handles_corrupt_large_index_count --quiet
- cargo test -p lix_engine tracked_state::context::tests --quiet
- git diff --check

Review note:
- Attempted the required three-agent review loop, but all spawned reviewers failed due to the account usage limit. This draft keeps the stack moving while making that review gap explicit.


samuelstroschein:

PR #465, opened 2026-05-19
agent-1: reject key-only binary CAS values

samuelstroschein:
Summary
- fail closed when binary CAS point reads request full values but storage returns key-only projected entries
- fail closed when binary CAS manifest chunk scans return key-only entries instead of silently skipping them
- add fake backend regressions for manifest point reads and manifest-chunk scans

Tests
- cargo test -p lix_engine load_manifest_rejects_key_only_point_read --quiet
- cargo test -p lix_engine scan_manifest_chunks_rejects_key_only_value --quiet
- cargo test -p lix_engine binary_cas::kv::tests --quiet
- git diff --check

Review loop
- Attempted 3 sub-agent reviews after tests passed.
- Initial attempt hit the local agent thread limit, so stale errored agents were closed and the review attempt was retried.
- All 3 retried review attempts were blocked by the Codex usage limit: try again at 7:12 AM.
- No reviewer feedback could be collected or applied for this PR before opening.

### DIFF
--- a/packages/engine/src/binary_cas/kv.rs
+++ b/packages/engine/src/binary_cas/kv.rs
@@ -152,7 +152,8 @@ async fn get_one(
         .into_iter()
         .next()
         .flatten()
-        .and_then(full_value))
+        .map(|value| full_value(space, value))
+        .transpose()?)
 }
 
 async fn scan_all_values(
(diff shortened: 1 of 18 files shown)


samuelstroschein:

PR #505, opened 2026-06-08
FsBackend: sync lix to/from a directory

alecmocatta:
```rust
let lix = open_lix_with_backend(FsBackend::open("./workspace").await?).await?;

lix.write_file("/hello.txt", b"hello".to_vec(), FsWriteOptions::default()).await?;
assert_eq!(std::fs::read("./workspace/hello.txt")?, b"hello");

lix.close().await?;
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2032,6 +2032,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
+name = "file-id"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1fc6a637b6dc58414714eddd9170ff187ecb0933d4c7024d1abbd23a3cc26e9"
+dependencies = [
+ "windows-sys 0.60.2",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


alecmocatta:
- **P1** Dirty disk edits can be overwritten by unrelated Lix writes. `materialize_snapshot()` preserves local-only files, but overwrites existing paths whenever bytes differ, without checking whether disk still matches `last_materialized`: [filesystem.rs:493](packages/rs-sdk/src/filesystem.rs:493), [filesystem.rs:546](packages/rs-sdk/src/filesystem.rs:546). A pending local edit to `/a.txt` can be reverted by an SDK write to `/b.txt` before the watcher imports `/a.txt`.

- **P1** `commit()` can report failure after the DB commit is already durable. [filesystem.rs:254](packages/rs-sdk/src/filesystem.rs:254) commits `inner` first, then syncs to disk. Any later materialization failure, for example local-only children blocking `remove_dir()` at [filesystem.rs:519](packages/rs-sdk/src/filesystem.rs:519), returns a backend commit error even though storage already advanced.

- **P1** The root symlink rejection is bypassed. The code opens `dir/.lix` before root validation at [filesystem.rs:870](packages/rs-sdk/src/filesystem.rs:870), then canonicalizes the root at [filesystem.rs:274](packages/rs-sdk/src/filesystem.rs:274), so the later `symlink_metadata` check at [filesystem.rs:659](packages/rs-sdk/src/filesystem.rs:659) sees the target directory, not the caller-provided symlink.

- **P1** `FsBackend` can fail to open a plugin-enabled `.lix` DB if `.lix_system/` is not already materialized on disk. Disk-to-Lix deletion skips `/.lix_system/plugins...` but not the parent `/.lix_system/`: [filesystem.rs:431](packages/rs-sdk/src/filesystem.rs:431), [filesystem.rs:803](packages/rs-sdk/src/filesystem.rs:803). Removing the parent then hits the engine’s recursive reserved-path rejection at [fs.rs:416](packages/engine/src/session/fs.rs:416).

- **P2** Runtime configuration is split across two independent engines. `FsBackend::open` creates the supervisor engine without the `OpenLixOptions::wasm_runtime` passed later to `open_lix`: [filesystem.rs:175](packages/rs-sdk/src/filesystem.rs:175), [lix.rs:63](packages/rs-sdk/src/lix.rs:63). Disk-originated plugin work can use a different runtime than foreground SDK operations.

- **P2** Plugin archive disk import is not idempotent. Every local file under `/.lix_system/plugins/` calls `install_plugin_archive()` without checking whether Lix already has identical bytes: [filesystem.rs:469](packages/rs-sdk/src/filesystem.rs:469). Unrelated disk edits can reinstall already-materialized archives and create avoidable history churn.

PR #521, opened 2026-06-13
Simplify Lix path handling

alecmocatta:
Fixes:

- Fixes over-strict path canonicalization. Old `fs_path` treated Lix paths like URI/PRECIS identities; the branch makes path segments opaque text with only slash/empty/`.`/`..` structural rules. This preserves `%`, `#`, long segments, decomposed Unicode, zero-width text, trailing-dot names, `.env`, etc. See [lix_path.rs](packages/engine/src/common/lix_path.rs:3), [lix_file.rs](packages/engine/tests/sql/lix_file.rs:172), [lix_directory.rs](packages/engine/tests/sql/lix_directory.rs:64).

- Makes SQL `path` behave like normal text. Removed special predicate normalization that decoded/canonicalized path filters and rejected computed predicates; `lower(path)` now works, and percent/Unicode spellings compare literally. See [lix_file.rs](packages/engine/tests/sql/lix_file.rs:7) and [datafusion.rs](packages/engine/src/sql2/exec/datafusion.rs:2084).

- Moves descriptor-name validation to the right boundary. Transaction normalization no longer mutates filesystem descriptor names; schema/rendering still reject `/`, but descriptor names are otherwise opaque. This avoids collisions/mutations like decomposed `Café` becoming `Café`. See [transaction/normalization.rs](packages/engine/src/transaction/normalization.rs:517) and [lix_directory_descriptor.json](packages/engine/src/schema/builtin/lix_directory_descriptor.json:35).

- Fixes descriptor-graph path planning. Directory/file paths are derived from descriptor parent/name graphs, not cached normalized strings, so descriptor-shaped writes can still materialize/read correctly. See [descriptor_path.rs](packages/engine/src/filesystem/descriptor_path.rs:17) and [planner.rs](packages/engine/src/filesystem/planner.rs:944).

- Refactors `SessionFs` onto SQL filesystem views. `fs.write_file`, `read_file`, `mkdir`, `readdir`, and `rm` now use `lix_file` / `lix_directory`, reducing duplicated filesystem semantics between API and SQL. See [session/fs.rs](packages/engine/src/session/fs.rs:65).

- Fixes tracked/untracked path interaction. Broad cross-lane ancestor rejection is replaced with more useful promotion/reuse: tracked writes can promote untracked parents, untracked children can reuse tracked parents, exact same-path shadowing still conflicts. See [lix_directory.rs](packages/engine/tests/sql/lix_directory.rs:980), [lix_file.rs](packages/engine/tests/sql/lix_file.rs:1988).

- Makes `FsBackend` ignore unmanaged disk entries safely. Symlink roots still fail, but symlink children, sockets, invalid UTF-8 names, and SQLite sidecars are ignored instead of imported/failing; materialization avoids overwriting or following unmanaged blockers. See [filesystem.rs](packages/rs-sdk/src/filesystem.rs:678) and [e2e.rs](packages/rs-sdk/tests/e2e.rs:636).

- Regularizes plugin archive storage. Plugin install is now modeled as writing a reserved archive file, with manifest-key/path checks and guarded reserved-path mutation/delete behavior. Archive paths are parsed with explicit slash-only portable rules instead of `std::path::Path`. See [plugin/storage.rs](packages/engine/src/plugin/storage.rs:33), [archive.rs](packages/engine/src/plugin/archive.rs:304), [fs_api.rs](packages/engine/tests/fs_api.rs:408).

- Improves plugin diff/reconciliation context. File-data reconciliation no longer assumes every file-data write has a path, plugin storage writes are separated from normal plugin detection, and plugins receive `filename` alongside bytes without exposing full paths. See [transaction/context.rs](packages/engine/src/transaction/context.rs:480), [lix-plugin.wit](packages/engine/wit/lix-plugin.wit:7).


alecmocatta:
@codex review

alecmocatta:
@codex review

PR #610, opened 2026-07-17
Decouple Lix facade from worker transport

samuelstroschein:
## What changed

- make the public `Lix` facade depend on the semantic `LixBinding` boundary
- adapt the existing worker client behind that binding
- move transaction and observation lifecycle handling off worker-specific IDs

## Why

`Lix` was coupled directly to its local worker message protocol, forcing remote consumers to reimplement the public client. This creates the transport-neutral seam needed for a first-class remote HTTP binding without changing local behavior.

## Impact

There is no intended public API or runtime behavior change. Local memory, SQLite, filesystem, transaction, branch, and observation operations continue to use the existing worker transport through an adapter.

## Validation

- `npm run build`
- `npm run typecheck`
- `npx vitest run src/open-lix.test.ts` (65 tests)

### DIFF
--- a/packages/js-sdk/src/open-lix.ts
+++ b/packages/js-sdk/src/open-lix.ts
@@ -4,12 +4,14 @@ import {
 	invalidArgument,
 } from "./errors.js";
 import type {
-	BindingExecuteResult,
 	BindingObserveEvent,
+	LixBinding,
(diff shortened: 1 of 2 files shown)


samuelstroschein:

PR #611, opened 2026-07-17
Add unified remote Lix HTTP mode

samuelstroschein:
## What changed

- add unified `openLix({ server: { mode: "remote", url } })` support
- define the versioned `/.lix/v1/` handshake, execute, and branch contract in Lix
- support dynamic headers and an injected `fetch` for browser and service-binding callers
- serialize remote operations to preserve normal Lix call ordering
- lazy-load the local worker path only when local mode is selected
- expose the pure remote protocol as `@lix-js/sdk/remote-protocol`

## Why

Lix consumers should use the normal client against a server instead of maintaining application-specific HTTP facades. This is the finite-operation layer; remote observation is the next stacked PR.

Remote mode is deliberately thin and does not use client storage. A local `storage` plus a server belongs to the future sync mode rather than this MVP.

## Validation

- `npx vitest run src/remote/client.test.ts src/open-lix.test.ts` (72 tests)
- `npm run build:ts`
- `npm run typecheck`
- `git diff --check`

## Stack

- base: #610
- next: remote `lix.observe()`

### DIFF
--- a/.changenotes/add-remote-lix-client.md
+++ b/.changenotes/add-remote-lix-client.md
@@ -1,0 +1,7 @@
+---
+type: minor
+---
+
+Lix can now open a workspace as a thin remote client.
+
+Use `openLix({ server: { mode: "remote", url } })` to execute SQL and manage branches through the versioned Lix HTTP protocol without loading a local engine.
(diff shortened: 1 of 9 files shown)


samuelstroschein:

PR #613, opened 2026-07-18
Add a remote-only Lix entrypoint

samuelstroschein:
## What changed

- adds `@lix-js/sdk/remote`, exposing the normal `openLix`/`Lix` facade for thin clients
- separates the transport-neutral facade from local storage and worker loading
- documents the deployment-sensitive entrypoint
- extends the packed Vite smoke test to assert that the remote entrypoint emits neither the browser worker nor WASM assets

## Why

The unified `openLix({ server: { mode: "remote", url } })` API correctly avoids opening a local engine at runtime, but bundlers still discovered its dynamic local-worker import and emitted the 37.5 MiB engine WASM. That exceeds Cloudflare's 25 MiB per-asset limit. The published release WASM is already compiled for size and symbol stripping saved only 74 bytes, so remote-only deployments need a package graph that cannot reach local-engine modules.

This keeps the normal unified entrypoint for mixed local/remote applications while giving thin-client deployments the same facade through a hard no-WASM boundary.

## Validation

- `npm run typecheck`
- `npx vitest run` (86 tests)
- `npm run test:browser:production`
- remote packed Vite output: one 23.47 kB JavaScript asset, zero worker or WASM assets
- `git diff --check`


samuelstroschein:

PR #649, opened 2026-07-21
perf(engine): bypass DataFusion for exact primary-key reads

samuelstroschein:
## Hypothesis

The common exact-primary-key `SELECT` path pays DataFusion's catalog registration, logical planning, optimization, and physical planning costs even though Lix providers already know how to load exact file IDs and entity PKs. A strict native binder should remove that fixed CPU cost without changing storage behavior.

This PR is independent and based directly on `origin/main` at `78398a8b`.

## What changed

- Add a pure `Declined | Ready` binder for one active `lix_file` or typed entity surface.
- Accept only explicit column projections and AND trees of `=` / non-negated `IN` over the complete primary key.
- Support reversed equality, text literals, `$N` text parameters, intersections, deduplication, aliases, and flat-string composite PKs up to 4,096 expanded keys.
- Reuse one prepared public catalog and cached branch reader for native execution or DataFusion fallback.
- Reuse the existing file/entity provider `plan_scan` and `load` implementations, preserving branch validation, global overlays, path-index selection, blob reads, plugin rendering, and Arrow metadata.
- Preserve JSON field metadata through aliases.
- Keep transaction-overlay reads on DataFusion.
- Make fallback possible only before provider I/O; loader errors never trigger a second execution.
- Include an ignored, release-only benchmark with deterministic 10k-row snapshot export/import, raw samples, p50/p95, storage counters, build provenance, and gated profiling markers.

Complex SQL, partial/non-string/nested PKs, non-text parameters, contradictions, wildcard/expression projections, OR/residual predicates, by-branch/history surfaces, joins, CTEs, ordering, limits, grouping, and oversized products deliberately remain on DataFusion.

## Release evidence

CPU 4, in-memory storage, exact 7,925,508-byte snapshot, BLAKE3 `56b43e134ba610d7c5e7cae9daa799d71f914b7f1c90c1b0d02ecb7be2a20308`, 10 warmups, 100 raw samples/case. Baseline `0040e220` contains only the benchmark; candidate is `df9fc9e4`.

| Case | Baseline p50 / p95 | Candidate p50 / p95 | Speedup p50 / p95 |
| --- | ---: | ---: | ---: |
| Entity `=` | 3.091 / 3.201 ms | 2.589 / 2.723 ms | 1.194x / 1.175x |
| Entity `IN 10` | 3.639 / 3.796 ms | 3.080 / 3.225 ms | 1.182x / 1.177x |
| Entity `IN 100` | 8.504 / 8.933 ms | 7.473 / 7.703 ms | 1.138x / 1.160x |
| File descriptor `=` | 1.810 / 1.901 ms | 1.054 / 1.131 ms | 1.718x / 1.681x |
| File descriptor `IN 10` | 2.156 / 2.253 ms | 1.335 / 1.434 ms | 1.616x / 1.572x |
| File descriptor `IN 100` | 3.095 / 3.213 ms | 1.864 / 1.962 ms | 1.660x / 1.637x |
| File data `=` | 1.939 / 2.056 ms | 1.171 / 1.251 ms | 1.656x / 1.644x |
| File data `IN 10` | 3.157 / 3.314 ms | 2.306 / 2.478 ms | 1.369x / 1.337x |
| File data `IN 100` | 10.814 / 11.321 ms | 9.832 / 10.252 ms | 1.100x / 1.104x |

All nine storage-counter distributions are identical before/after. For entity equality that is 2 `begin_read`, 27 `get_many` calls / 86 requested keys, and 4 scans per query. The narrowing gain at 100 blobs shows the intended fixed-cost behavior rather than hidden I/O reduction.

## Profile evidence

Matched `perf record -e cycles:u -F 997 -g --call-graph dwarf,16384`, 4,000 entity-equality queries, explicit ready/go/done/ack boundaries, zero lost samples:

- sampled cycles/query: 9.559M -> 7.914M (`-17.2%`)
- profiled loop: 13.253 s -> 11.071 s (`-16.5%`)
- storage totals remained exactly 8,000 reads, 108,000 `get_many` calls / 344,000 keys, and 16,000 scans

The candidate flame graph is correspondingly more concentrated in tracked-state decoding/traversal and allocation: the useful provider/storage work is now the floor.

## Validation

- `cargo clippy -p lix_engine --lib -- -D warnings`
- 8/8 native binder/executor tests
- native-vs-forced-DataFusion differential test for file descriptors/blobs and composite entity PKs/aliased JSON
- public-path nine-case smoke
- `cargo test -p lix_engine`: full package green, including 1,022 unit tests, 546 SQL simulation/conformance tests, storage/transaction tests, and doctests
- matched release benchmark and cycle profiles above

## Deliberate 10% exclusions

- no direct storage loader in v1; the existing provider path remains the semantic authority
- no native transaction-overlay reads
- no alternate SQL string literal syntaxes beyond ordinary single-quoted text
- no attempt to cache catalogs or prepared SQL across calls in this PR
- no native empty-intersection result, preserving current active-branch/error ordering through fallback

### DIFF
--- a/packages/engine/src/session/execute.rs
+++ b/packages/engine/src/session/execute.rs
@@ -1232,4 +1232,107 @@ mod tests {
             serde_json::json!("value")
         );
     }
+
+    #[tokio::test]
+    async fn exact_primary_key_reads_match_datafusion_fallback_results() {
+        let session = open_session().await;
+        session
+            .execute(
+                "INSERT INTO lix_file (id, path, data) VALUES \
+                 ('native-file-a', '/native-a.bin', X'6100'), \
+                 ('native-file-b', '/native-b.bin', X'6200')",
+                &[],
+            )
+            .await
+            .expect("file fixtures should insert");
+
+        let file_params = [
+            Value::Text("native-file-b".to_string()),
+            Value::Text("native-file-a".to_string()),
+        ];
+        let native_files = session
+            .execute(
+                "SELECT id AS file_id, path, data AS bytes \
+                 FROM lix_file WHERE id IN ($1, $2)",
+                &file_params,
+            )
+            .await
+            .expect("native file read should succeed");
+        // The tautology is deliberately outside the native binder's exact-key
+        // language, forcing the same projection through DataFusion.
+        let datafusion_files = session
+            .execute(
+                "SELECT id AS file_id, path, data AS bytes \
+                 FROM lix_file WHERE id IN ($1, $2) AND id = id",
+                &file_params,
+            )
+            .await
+            .expect("DataFusion file read should succeed");
+        assert_eq!(native_files, datafusion_files);
+
+        let schema = serde_json::json!({
+            "x-lix-key": "native_pk_message",
+            "x-lix-primary-key": ["/workspace_id", "/id"],
+            "type": "object",
+            "properties": {
+                "workspace_id": { "type": "string" },
+                "id": { "type": "string" },
+                "payload": { "type": "object" }
+            },
+            "required": ["workspace_id", "id", "payload"],
+            "additionalProperties": false
+        });
+        session
+            .execute(
+                &format!(
+                    "INSERT INTO lix_registered_schema \
+                     (value, lixcol_global, lixcol_untracked) \
+                     VALUES (lix_json('{}'), false, false)",
+                    schema
+                ),
+                &[],
+            )
+            .await
+            .expect("entity schema should register");
+        session
+            .execute(
+                "INSERT INTO native_pk_message (workspace_id, id, payload) VALUES \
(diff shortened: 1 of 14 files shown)


samuelstroschein:
Closing this implementation because it proves a real ~0.5–1.2 ms fixed planning cost, but addresses it by adding a second SQL binder/executor. Storage work was unchanged, so the follow-up investigation will first measure and optimize reusable DataFusion session, catalog, provider, and parameterized logical-plan state.

#649 remains the measured lower-bound/reference implementation. We should only revisit a specialized SQL execution path if the standard DataFusion approaches cannot approach these results and measurements show that the additional semantic and maintenance cost is justified. The follow-up should start from current main, which now includes #648 and #651.

PR #694, opened 2026-07-22
perf(engine): fast path key-value point reads

samuelstroschein:
## Summary

- recognize Atelier's exact parameterized observer shape: `SELECT value FROM lix_key_value WHERE key = $1`
- lower that shape directly to the live-state exact reader instead of constructing a DataFusion catalog and plan
- preserve active-branch-over-global visibility, tombstones, JSON/null conversion, missing rows, and branch validation
- conservatively retain the generic SQL path for aliases, literals, extra projections/predicates, joins, limits, and every other shape

This is crate-private and does not change the public API.

## Results

Five alternating-order process pairs, pinned to CPU 7, with 50 warmups and 1,000 measured iterations per case. The table reports the median of run-level p50/p95 latency. Baseline and candidate were built from the same engine tree at `35f7e1c691fb4108a7d20c392fe2c18fb2d6c676`; current `main` only adds CI/JS-build changes in #692.

| Backend / case | p50 before → after | p50 | p95 before → after | p95 |
| --- | ---: | ---: | ---: | ---: |
| RocksDB active row | 573,949 → 93,556 ns | **-83.7%** | 615,397 → 119,736 ns | **-80.5%** |
| RocksDB global fallback | 575,952 → 96,091 ns | **-83.3%** | 614,405 → 125,946 ns | **-79.5%** |
| RocksDB branch override | 576,623 → 98,816 ns | **-82.9%** | 615,647 → 124,845 ns | **-79.7%** |
| RocksDB global session | 515,409 → 41,849 ns | **-91.9%** | 553,070 → 58,609 ns | **-89.4%** |
| RocksDB missing key | 564,702 → 88,356 ns | **-84.4%** | 611,361 → 113,745 ns | **-81.4%** |
| SlateDB active row | 987,147 → 425,711 ns | **-56.9%** | 1,055,475 → 477,649 ns | **-54.7%** |
| SlateDB global fallback | 986,506 → 432,994 ns | **-56.1%** | 1,056,017 → 501,092 ns | **-52.5%** |
| SlateDB branch override | 1,006,173 → 458,742 ns | **-54.4%** | 1,086,854 → 612,141 ns | **-43.7%** |
| SlateDB global session | 738,209 → 213,431 ns | **-71.1%** | 787,041 → 260,821 ns | **-66.9%** |
| SlateDB missing key | 959,385 → 403,908 ns | **-57.9%** | 1,034,415 → 461,909 ns | **-55.3%** |

The deliberately non-matching aliased-query control stayed within noise: RocksDB +1.3% p50/+0.2% p95; SlateDB +0.8%/+3.0%. Deterministic storage traces show that matching reads replace one or two scans with correlated exact batches; control traces are unchanged.

Artifacts:

- identical harness SHA-256: `18967354b354a3f234479cad6dd50fc28e959bf95483992f0738f54ced935375`
- baseline binary SHA-256: `664cf9005bb3ce3b414be0671e36dd1f11b82e6ddc1f1165a7f0ac8af60c57cb`
- candidate binary SHA-256: `6a4164c54c0ebecc73f5f404d2c7963bdd937709b519336b1918a4cf3452de90`

The manual benchmark harness is intentionally excluded from this PR.

## Why this shape

This follows the same principle as DataFusion's exact filter pushdown and DuckDB's primary-key point probes: use a selective equality predicate to avoid scanning and generic relational machinery. It also maps naturally to SlateDB's native key-oriented read path.

- https://datafusion.apache.org/library-user-guide/custom-table-providers.html#filter-pushdown-doing-less-work
- https://duckdb.org/docs/lts/guides/performance/indexing
- https://slatedb.io/docs/design/reads/

The recognizer is intentionally narrow. Unsupported syntax falls back before any semantic work, keeping the optimized 90% path bounded and maintainable.

## Validation

- `cargo test -p lix_engine --lib`: 1,142 passed, 0 failed, 6 ignored
- focused classifier: 1 passed
- focused storage-backed SQL simulation modes: 2 passed
- `cargo clippy -p lix_engine --all-targets --all-features -- -D warnings`
- `git diff --check`
- independent static review: no actionable findings

### DIFF
--- a/.changenotes/fast-key-value-point-read.md
+++ b/.changenotes/fast-key-value-point-read.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+Improved parameterized `lix_key_value` observer reads by resolving exact keys
+through the live-state point reader instead of constructing a DataFusion plan.
+Active-branch overrides and global fallback retain their existing behavior.
--- a/packages/engine/src/session/execute.rs
+++ b/packages/engine/src/session/execute.rs
@@ -485,6 +485,7 @@ where
 
         let acknowledge_file_views = is_acknowledgeable_file_data_read(&statement, params);
         let exact_lix_file_read = exact_lix_file_read(&statement, params);
+        let exact_lix_key_value_read = exact_lix_key_value_read(&statement, params);
         let has_durable_runtime_function = sql2::statement_has_durable_runtime_function(&statement);
         let runtime_write_access = if has_durable_runtime_function {
             let write_access = self.begin_session_write_access().await?;
@@ -520,6 +521,7 @@ where
                     params,
                     acknowledge_file_views,
                     exact_lix_file_read,
+                    exact_lix_key_value_read,
                     has_durable_runtime_function,
                 )
                 .await
@@ -977,6 +979,7 @@ where
         params: &[Value],
         acknowledge_file_views: bool,
         exact_lix_file_read: Option<(sql2::ExactLixFileReadSelector, sql2::ExactLixFileReadColumn)>,
+        exact_lix_key_value_read: Option<String>,
         has_durable_runtime_function: bool,
     ) -> Result<
         (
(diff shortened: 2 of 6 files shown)


samuelstroschein:
Closing after broader design review. This raw-AST shortcut would create a table-specific secondary planning path, and it is not universally SQL-equivalent: `x-lix-primary-key` alone is not the full storage identity because entity rows may be file-scoped, while this implementation hard-codes `file_id: None`. A future candidate must be generalized from DataFusion's bound logical plan and require a provider-confirmed complete identity (including file/branch scope) before direct execution; all other shapes must fall back.

PR #697, opened 2026-07-22
research: design an incremental Wasm file plugin API

samuelstroschein:
## Summary

This is PR1 of the incremental Wasm file-plugin work. It adds the research/RFC,
compileable API and WIT sketches, executable Wasm mechanism probes, controlled
AX evaluations, and reproducible RocksDB/SlateDB discovery baselines. It does
not change default file semantics.

The proposed author API keeps plugins in Wasm and has four lifecycle calls:

- `open_file`: raw bytes to a persistent private document plus initial groups
- `open_entities`: durable entities to a shared renderer plus complete bytes
- `file_changed`: byte splices to a successor document plus semantic groups
- `entities_changed`: resolved groups/activation hydration to a successor
  renderer plus byte patches

Generated IDs use one compact retry-stable namespace passed inline with a file
transition. The SDK derives one 32-character PK component locally, so a 200k-row
import makes zero allocator imports. Parentage and order remain entity fields,
which preserves unambiguous cross-parent moves. Existing UUID PKs are preserved
during migration.

## Evidence

The retained exact-`e1a57ec3` current-v1 run uses the real CSV Wasm plugin and
ordinary SQL blob writes over RocksDB `LocalFilesystem` and cached SlateDB:

- A one-row edit requested 226,318 / 226,283 keys respectively.
- Both lanes requested 220,092 change-payload keys and returned 220,089 values.
- RocksDB edit p50/p95 was 2,568/2,718 ms; SlateDB was 3,836/4,412 ms.
- Exact render p50/p95 was 855/960 ms and 4,080/4,374 ms.
- The classifier was corrected to use exact blocking leaves; retained profiles
  report both total and active views.

The isolated B2 mechanism is 264.9-1462.6x faster than its deliberately
optimistic v1 mechanism control, but that is a guest-side lower bound, not a
Lix write-speed claim. PR2 must demonstrate the end-to-end result.

Other tested hypotheses include persistence-only blobs, splice/patch documents,
host byte sources, copied checkpoints, host KV context, a packed transient ABI,
and WASI 0.3/P3 streams. The report records latency, boundary bytes, guest high-
water, storage/RSS limits, and rejected alternatives.

Controlled AX results were A=5 (91.6 mean), B=9 (90.3), C=4 (91.5), and D=4
(87.5), followed by a heterogeneous refined N=3 (86/85/82) and one final-
aligned score of 87. Post-evaluation correctness changes—activation hydration,
compact inline IDs, packet ordering, and generated-binding details—are clearly
marked as requiring a pre-freeze AX follow-up.

## Acceptance boundary

PR2 is stacked on this branch and owns the production CSV vertical slice. It
must eliminate the 220,092 requested / 220,089 returned payload materializations
on the warm path, run under the 64 MiB guest cap, and clear a preregistered
greater-than-20% p50 **and** p95 paired gate independently on RocksDB filesystem
and cached SlateDB. It must also preserve multiplayer/acknowledgement semantics,
legacy IDs, cold render equality, and actor failure behavior.

The SlateDB discovery lane uses a local object store with production cache
budgets; remote-object-store latency/miss/retry/tail validation remains a
separate Lixray production gate. SQLite was not used for performance evidence.
No generic packed-KV storage layer is proposed.

`origin/main` advanced to `4aac696c` after the immutable e1 capture. Its
`35f7e1c6` parent moves result scalar payloads and can affect blob-return/render
cost, so PR2 must rebuild both paired arms from the same then-current main base.

## Validation

- Candidate facade: 11/11 tests, rustfmt, Clippy with warnings denied
- AX harness: 6/6 tests; all result JSON is schema-valid
- WIT: parse, binary encode, validation, and round-trip parse
- Mechanism/ABI/P3 probes and retained profile bundles validated
- Changed-document links: no broken local links
- `git diff --check`

The full report is in
[`perf-results/plugin-api-v2/plugin-api-v2-research-2026-07-22.md`](perf-results/plugin-api-v2/plugin-api-v2-research-2026-07-22.md),
and the proposed contract is in
[`rfcs/003-incremental-wasm-file-plugins/index.md`](rfcs/003-incremental-wasm-file-plugins/index.md).


samuelstroschein:
Superseded by #728, which consolidates the canonical production plugin v2 API and retained benchmark-proven implementation. Closing this preceding experiment/stack to keep review focused on #728.

PR #699, opened 2026-07-23
feat(protocol): add branch-pinned remote sessions

samuelstroschein:
## What

- add validated branch-pinned Lix sessions and expose them through the canonical handshake
- add JS SDK server.branchId support with fail-closed scope verification
- keep workspace sessions unchanged for clients that intentionally switch the shared branch
- reject branch switching on pinned handles and close unregistered or malformed sessions safely
- add protocol/SDK regressions, documentation, a changenote, and a permanent Criterion benchmark

## Why

Workspace sessions resolve the shared workspace branch selector on every operation and can be moved by another client. Agent-style clients normally stay on one branch, so a server-owned immutable scope removes both the selector read and the switch race.

## Performance

Release Criterion benchmark, 2 KiB exact-file reads, 50 samples:

- workspace selector: 45.895 us p50, 52.955 us p95
- branch pinned: 32.900 us p50, 34.152 us p95
- improvement: 28.3% p50 and 35.5% p95; Criterion point estimate 29.7%

The existing workspace path remains the in-process control and is otherwise unchanged.

## Safety and review

Independent review found and this branch fixed three issues before publication: invalid branch opens no longer evict a valid LRU session, the client verifies both reported scope and active branch, and malformed handshakes best-effort delete an issued session without masking the original error. A clean re-review found no remaining actionable issue.

## Verification

- Rust SDK: 12 passed, 1 ignored
- server protocol: 35 passed, 1 ignored
- remote JS client/protocol: 46 passed
- focused all-target/all-feature Clippy with warnings denied
- Rust formatting, diff check, and changenote validation

PR #702, opened 2026-07-23
bench: guard pinned session read performance

samuelstroschein:
## What

- add a paired engine benchmark for exact 2 KiB file reads through a workspace-selector session and a branch-pinned session
- add an in-process protocol benchmark that handshakes through the default remote path, retains the session ID, and repeatedly reads the same file through Axum
- verify the expected blob before timing so an empty or stale result cannot appear fast

## Why

#700 already contains the optimization explored in #699: a default remote handshake resolves the current branch once and opens a pinned engine session. Subsequent requests reuse that handle, while an explicit branch switch replaces it. This PR records the attributable performance difference and guards the optimized default protocol path without reintroducing the public immutable-branch API from #699.

## Results

Criterion point estimates on the same host, pinned to one CPU, with 5 s warm-up, 20 s measurement, and 100 samples:

| Benchmark | Workspace selector | Branch pinned | Change |
| --- | ---: | ---: | ---: |
| Core exact-file read | 48.359 µs | 35.588 µs | **-26.4%** |

The identical in-process protocol harness was also run in three alternating pairs on pre-#700 `defa42a9` and #700/main `9c0f9ce6`:

| Run | Pre-#700 | #700/main |
| --- | ---: | ---: |
| 1 | 84.714 µs | 59.557 µs |
| 2 | 85.278 µs | 62.458 µs |
| 3 | 84.776 µs | 58.698 µs |
| Median | 84.776 µs | 59.557 µs |

That is **29.7% lower** at the median of the run-level point estimates. Each pair improves by more than 10%.

Scope: the protocol benchmark uses memory storage and `Router::oneshot`. It includes handshake-derived session routing, JSON/Axum handling, leasing, and response-body consumption, but excludes sockets, TLS, the JS client, production storage, and concurrent server throughput. These are Criterion estimates, not p50/p95 latency claims.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `cargo clippy --profile test -p lix_sdk -p lix_server_protocol --all-targets --all-features -- -D warnings`
- `cargo test --profile test -p lix_sdk -p lix_server_protocol --all-features` (104 passed, 2 ignored)
- both benchmark targets compiled and executed successfully

No changenote: this is a repository benchmark-only change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3435,6 +3435,7 @@ version = "0.8.4"
 dependencies = [
  "async-stream",
  "axum",
+ "codspeed-criterion-compat",
  "flate2",
  "futures-core",
  "getrandom 0.3.4",
--- a/packages/rs-sdk/Cargo.toml
+++ b/packages/rs-sdk/Cargo.toml
@@ -79,3 +79,8 @@ name = "sqlite"
 path = "benches/sqlite.rs"
 harness = false
 required-features = ["sqlite"]
+
+[[bench]]
+name = "branch_session"
+path = "benches/branch_session.rs"
+harness = false
--- a/packages/rs-sdk/benches/branch_session.rs
+++ b/packages/rs-sdk/benches/branch_session.rs
@@ -1,0 +1,63 @@
+use std::hint::black_box;
+
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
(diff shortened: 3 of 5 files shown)


samuelstroschein:
Closing because #700 already contains the production pinned-session optimization, and this PR only adds benchmark coverage. The benchmark-only change is not needed for the current optimization work.

PR #720, opened 2026-07-23
Add recursive incremental JSON Component v2 plugin

samuelstroschein:
## Summary

Adds the production recursive JSON Component v2 vertical slice on top of the
incremental CSV v2 API.

- models every JSON node as a recursive semantic entity, including nested
  object members and stable opaque array-item IDs
- preserves decoded JSON Pointer identity for object keys
- keeps immutable persistent documents and emits one sparse semantic change
  for a one-leaf edit
- adds exact 10 MB flat and nested fixtures plus resumable RocksDB/SlateDB
  benchmark campaigns

This is the base of a stacked optimization series. It intentionally does not
add a public authoring SDK or change the production v2 WIT.

## Data

Production RocksDB filesystem, exact 10,000,000-byte JSON, 220,000 leaf
properties, one middle leaf changed:

| Shape | Edit p50 | Edit p95 | Guest high-water |
| --- | ---: | ---: | ---: |
| flat | 70.455 ms | 72.255 ms | 160,169,984 B |
| nested | 70.029 ms | 84.674 ms | 202,833,920 B |

Every warm edit emitted one durable semantic change and performed zero source
reads, full semantic materializations, full reparses, or filesystem full
renders. The prior v1-shaped full reopen was measured in seconds on the same
10 MB class of fixture; v2 makes work proportional to the edited leaf.

## Authorship

The public v2 WIT is unchanged from the production TSV reference evaluated by
the retained N=10 AX cohort (10/10 success, median score 76). This PR adds a
real recursive format implementation and executable reference without
prematurely introducing a facade.

## Validation

- 21 native recursive JSON tests
- production Wasm Component build and archive install
- production RocksDB filesystem setup/edit benchmark
- paired campaign/gate unit tests

Stack base: #701

PR #721, opened 2026-07-23
Skip redundant v2 rendering for uncontended writes

samuelstroschein:
## Summary

First optimization stacked on #720.

When the actor lease and durable semantic root prove that an acknowledged
observation is current, the engine now adopts the already validated
`file-changed` successor directly. It no longer sends the same sparse change
back through `entities-changed` to reproduce identical bytes.

Stale observations retain the renderer path, so concurrent edits still
compose against the accepted document.

## Benchmark decision

Production RocksDB filesystem, exact 10,000,000-byte JSON, 220,000 leaves,
sparse splice provenance:

| Deterministic work | Baseline | Candidate |
| --- | ---: | ---: |
| packet pages / records | 2 / 2 | 1 / 1 |
| component imports | 1 | 0 |
| flat boundary bytes | 458 B | 204 B |
| nested boundary bytes | 622 B | 286 B |

This cuts packets and records by 50%, component imports by 100%, and boundary
traffic by 54–56%. Across 36 measurements per shape, end-to-end RocksDB
latency remained storage-bound and effectively neutral; the report does not
claim a latency win.

All rounds retained one durable semantic change and zero source reads, full
semantic materializations, full reparses, full renderer invocations, and
filesystem full renders.

## Authorship

No public WIT, packet contract, or plugin implementation changes. The existing
N=10 v2 AX result therefore remains the applicable authoring baseline.

## Validation

- `cargo check -p lix_engine --lib`
- `cargo test -p plugin_json_incremental_v2 --lib`
- engine benchmark gate Python tests
- flat and nested production RocksDB benchmark
- `git diff --check`

Full samples and interpretation are retained in
`perf-results/plugin-api-v2/json-v2-pareto-stack-2026-07-23.md`.

### DIFF
--- a/experiments/plugin-api-v2/engine-bench/README.md
+++ b/experiments/plugin-api-v2/engine-bench/README.md
@@ -54,9 +54,10 @@ Every measured Component v2 edit (CSV or the JSON v2 arm) resets and prints the
 engine's aggregate transition counters as `plugin_v2_counters`. The harness
 fails the run unless the warm single-row or single-property edit materializes
 no full semantic state, requests and returns fewer than 64 change payloads,
-persists exactly one semantic change, hits both the private actor and shared
-renderer documents once, and performs no full document reparse, full renderer
-invocation, or filesystem-sync full render. These are work invariants in
+persists exactly one semantic change, hits the private actor document once,
+skips the shared renderer for an uncontended current-base edit, and performs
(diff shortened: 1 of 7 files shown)


samuelstroschein:
Superseded by #728, which consolidates the canonical production plugin v2 API and retained benchmark-proven implementation. Closing this preceding experiment/stack to keep review focused on #728.

PR #728, opened 2026-07-24
feat(plugin): add canonical incremental Wasm Component API v2

samuelstroschein:
## Outcome

This is the hard cut to the canonical incremental Wasm Component plugin API. It removes the v1 runtime and v1 plugins rather than carrying migration or matcher compatibility.

The API remains sandboxed Wasm with a typed, versioned WIT boundary. A persistent per-file actor accepts bounded byte reads or semantic entities and returns sparse byte/entity changes through:

- `open-file`
- `open-entities`
- `file-changed`
- `entities-changed`

The engine provides rollback-safe actor publication, durable mutation-scoped ID namespaces, exact splice provenance, a hard workspace-wide live-Store admission budget, and a configurable 128 MiB default guest-memory limit.

Canonical reference plugins cover four different shapes:

- recursive JSON Pointer entities
- CSV/TSV tables and rows
- GitHub Flavored Markdown structure
- Excalidraw scenes, elements, and files

JSON uses deterministic last-writer-wins lifecycle behavior for stale whole-file writes until conflicts become a first-class API. No Preview 3 API or parallel v1 surface remains.

The normative contract and author quickstart are in `packages/engine/wit/v2`; RFC 003 records the design.

## Evidence

The 10,680,000-byte, 220,000-row CSV fixture established the original v1-to-v2 result:

| Backend | v1 edit p50 | v2 edit p50 | v1 render p50 | v2 render p50 |
| --- | ---: | ---: | ---: | ---: |
| RocksDB | 6,507.439 ms | 63.610 ms | 2,317.470 ms | 18.013 ms |
| cached SlateDB | 9,659.544 ms | 80.184 ms | 7,600.187 ms | 6.397 ms |

The recursive JSON plugin gates both flat and nested 10 MiB fixtures; a one-leaf edit emits one semantic change. A custom non-WIT Core Wasm ABI prototype failed the preregistered replacement threshold: cold paths were slower and hot wins stayed below the required 20% cross-shape/backend improvement, so this PR keeps WIT.

The WIT authorship evaluation produced 10/10 successful implementations with median score 76 (p25 72.75, p75 82.75). A thin helper SDK remains deliberately separate from the protocol.

## Stack

1. This PR: canonical API, engine integration, and four reference plugins
2. #754: proportional byte-edit runtime
3. #779: persistent sparse Markdown state
4. #795: sparse semantic merge history
5. #783: Markdown/JSON performance gates
6. #790: deferred transport SHA-256

## Validation

[...]

PR #751, opened 2026-07-24
perf(binary-cas): bound sparse-edit chunk rewrites

samuelstroschein:
## Summary

- cap newly written FastCDC chunks at 1 MiB
- retain the existing 4 MiB decode ceiling, so persisted manifests remain readable
- add a regression proving multi-megabyte low-entropy payloads cannot collapse into one rewritten chunk

## Root cause

The default profile targeted 1 MiB chunks but allowed a 4 MiB maximum. The 3.5 MiB Markdown benchmark therefore collapsed into one chunk. A one-byte edit recompressed and restaged the whole document, consuming about 8.9 ms per edit and defeating content-addressed reuse.

This follows the same bounded-work principle used by scalable storage systems: Sapling avoids O(repository) work through incremental state, DuckDB bounds storage work in blocks and row groups, and LSM systems separate small foreground writes from later compaction.

## Measured impact

Identical 3.5 MiB Markdown corpus, 7,253 paragraphs, 20 deterministic one-byte edits per cohort:

| cohort | Lix byte p50/p95 | Git p50/p95 |
| --- | --- | --- |
| 1 | 45.12 / 49.44 ms | 63.66 / 67.34 ms |
| 2 | 46.35 / 48.38 ms | 61.01 / 64.17 ms |
| 3 | 47.09 / 51.83 ms | 60.69 / 74.69 ms |

Lix is 22-29% faster at p50 and 24-31% faster at p95 in all three cohorts.

Cohort 1 incremental live storage after the identical edit trace:

- Lix: 11.77 MB
- Git before GC: 14.12 MB
- Git after GC: 3.96 MB

## Validation

- cargo test -p lix_engine binary_cas -- --nocapture (38 passed plus relevant integration test)
- cargo test -p lix_engine --features all-simulations (786 passed, 3 ignored)
- engine doc tests (3 passed)

### DIFF
--- a/packages/engine/src/binary_cas/chunking.rs
+++ b/packages/engine/src/binary_cas/chunking.rs
@@ -1,6 +1,7 @@
 pub(super) const INLINE_BINARY_CAS_MAX_BYTES: usize = 32 * 1024;
 pub(super) const SINGLE_CHUNK_FAST_PATH_MAX_BYTES: usize = 64 * 1024;
 pub(super) const MAX_BINARY_CAS_CHUNK_BYTES: usize = 4096 * 1024;
+const FASTCDC_WRITE_MAX_CHUNK_BYTES: usize = 1024 * 1024;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) struct BinaryCasChunking {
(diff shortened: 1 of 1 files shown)


samuelstroschein:
Superseded by the canonical hard-cut implementation in #728; the bounded CAS behavior is already included in that reviewed stack.

PR #754, opened 2026-07-24
perf(plugin): keep byte edits proportional to changed data

samuelstroschein:
Stacked on #728.

## Outcome

Keeps ordinary full-byte writes proportional to the changed region without changing SQL, WIT, or plugin author APIs.

- Reads of already-materialized plugin bytes no longer instantiate and hydrate a Wasm actor.
- Cold/blind complete-document writes use the documented last-writer-wins policy.
- Cold entity restore may reuse a host-verified accepted materialized base.
- CSV, JSON, Markdown, and Excalidraw implement the accepted-base contract.
- Fallback common-prefix/suffix discovery uses word-at-a-time comparisons.

This layer also preserves exact warm observations so unrelated concurrent semantic edits continue to compose.

## Current full-stack result

3,670,017-byte Markdown fixture, 7,253 paragraphs:

- Lix byte write + commit: 8.140 ms p50
- Git write + add + commit: 62.340 ms p50
- Lix semantic write: 7.609 ms p50

The cold-write path still reconstructs plugin state and is intentionally reported separately by the benchmark layer.

## Validation

- workspace clippy with `-D warnings`
- engine `all-simulations`
- four canonical plugin suites
- Markdown-vs-Git release benchmark

### DIFF
--- a/packages/engine/src/plugin/incremental.rs
+++ b/packages/engine/src/plugin/incremental.rs
@@ -285,7 +285,8 @@ pub(crate) fn build_file_update_splices(
                 0,
             )
         } else {
-            if before == after {
+            let (prefix, prefix_bytes_compared) = common_prefix_len(before, after);
+            if prefix == before.len() && prefix == after.len() {
                 return Ok(BuiltInputSplices {
                     edits: Vec::new(),
                     used_transport_provenance: false,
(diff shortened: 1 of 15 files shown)


samuelstroschein:
CI cleanup follow-up: the first cleaned SHA exposed a transactional self-deadlock. A staged `lix_file.data` read attempted to publish a cache observation while the same transaction held the actor lease; stale and absent views were also conflated. The updated single commit keeps staged observations transaction-private, preserves path/incarnation authority, and adds the corrected cold-read counter expectation. Exact full-workspace tests and all 21 active v2 SDK e2e tests pass locally.

PR #770, opened 2026-07-25
perf: group tracked serving head by entity key

samuelstroschein:
## What changed

Moves the tracked serving head to the v5 layout:

- one authoritative grouped record per `(branch, generation, schema, entity PK)` for normal tracked reads;
- a file-id-first member projection for both exact file identities and schema-scoped `file_id = ?` reads;
- a marker that atomically publishes the group generation with the branch head;
- repository protocol gate `tracked-head-group.v5`, so older physical layouts fail closed; and
- backend-neutral CAS preconditions for all branch-ref publication paths.

## Why

The normal tracked PK path previously had to prefix-scan rows because `file_id` was part of every physical row key. Grouping makes the normal known-PK path a point read, while the member projection prevents filesystem/file-ID reads from regressing or requiring sibling-group decoding.

Branch refs are the publication fence for the materialized head. The new `KeyAbsent` / `KeyValueEquals` guards prevent stale writers from publishing a ref move over a newer one on every storage backend.

## Compatibility

This is an intentional storage hard cut. Repositories initialized with the old protocol are rejected rather than interpreted with v5 visibility semantics.

## Validation

- `cargo test -p lix_engine tracked_head --lib`
- `cargo clippy -p lix_engine --features storage-benches -- -D warnings`
- `cargo test -p lix_rocksdb_storage`
- `cargo test -p lix_sqlite_storage`

## Before / after: release-mode hot CRUD

Same host, RocksDB backend, 10k seeded tracked rows, one warmed session, and
only the operation window timed. `origin/main` is the tree at `adc84a16`; the
PR is `00b52ef7`. Values are single-run hot-window means, so they are
directional rather than a statistical confidence interval.

| Operation | `main` | v5 PR | Change |
| --- | ---: | ---: | ---: |
| SQL session: read one tracked PK (100k repeats) | 35.704 us | 32.819 us | **-8.1%** |
| SQL session: read 10 tracked PKs (20k repeats) | 154.255 us | 153.609 us | -0.4% |
| SQL session: update one tracked PK (10k repeats) | 108.125 us | 106.227 us | -1.8% |
| Standalone raw SQLite: read one PK (100k repeats) | — | 2.696 us | v5 is 12.2x slower |
| Standalone raw SQLite: update one PK (10k repeats) | — | 5.459 us | v5 is 19.5x slower |

Reproduce the Lix values with:

```text
LIX_TRACKED_STATE_CRUD_PROFILE=1 \
LIX_TRACKED_STATE_CRUD_PROFILE_OP=read_one \
LIX_TRACKED_STATE_CRUD_PROFILE_LAYER=sql_session \
LIX_TRACKED_STATE_CRUD_PROFILE_HOT_REPEATS=100000 \
cargo bench -p lix_engine_benchmarks --bench tracked_state_crud --features storage-benches
```

## File-ID regression guard

`explicit_file_id_reads_use_single_member_projection` writes multiple
file-backed members, removes the grouped record, and verifies both exact
`(schema, PK, file_id)` lookup and schema-scoped `file_id = ?` scan still
return their rows. That proves these queries use the member projection rather
than fetch or unpack sibling groups.


samuelstroschein:
Added reproducible before/after data to the PR body: hot RocksDB SQL PK read is 35.704 us on main versus 32.819 us on v5 (-8.1%). It also now includes read-many, update, standalone SQLite comparisons, and the file-ID projection regression guard.

PR #771, opened 2026-07-25
perf: use point get for single RocksDB read

samuelstroschein:
## What changed

Routes a one-key `RocksDBRead::get_many` request through RocksDB's snapshot
`get` API instead of constructing and issuing a one-element `multi_get`.

## Why

The normal tracked v5 point-read path is increasingly dominated by individual
storage reads. A one-key batch has no batching benefit, so using the native
point API removes the vector/iterator setup on that path while retaining the
same snapshot and projection contract.

## Safety

The new focused test verifies an old snapshot still sees the pre-write value,
and covers full-value, key-only, and missing-key projections.

## Controlled release Criterion A/B

Same worktree, isolated target directory, release rebuild of
`lix_rocksdb_storage` for each side, RocksDB backend. This avoids reusing a
benchmark artifact built from the other source revision.

| One-key `get_many` case | Previous `multi_get` | This PR: `Snapshot::get` | Change |
| --- | ---: | ---: | ---: |
| Missing key, 100-row fixture | 1.0916 us | 0.7056 us | **-35.4%** (95% CI -36.0% to -33.9%) |
| 4 KiB hit | 1.8520 us | 1.2869 us | **-30.3%** (95% CI -31.8% to -28.1%) |

## Validation

- `cargo test -p lix_rocksdb_storage --test storage`
- `cargo clippy -p lix_rocksdb_storage -- -D warnings`

### DIFF
--- a/packages/rocksdb-storage/src/rocksdb.rs
+++ b/packages/rocksdb-storage/src/rocksdb.rs
@@ -262,6 +262,17 @@ impl StorageRead for RocksDBRead<'_> {
         opts: GetOptions,
     ) -> impl Future<Output = Result<GetManyResult, StorageError>> + Send {
         async move {
+            if let [key] = keys {
+                let physical_key = physical_key(space, key);
+                let value = self
+                    .snapshot
+                    .get(physical_key.0.as_ref())
+                    .map_err(rocksdb_error)?;
+                return Ok(GetManyResult::new(vec![
+                    value.map(|value| project_owned_value(value, opts.projection)),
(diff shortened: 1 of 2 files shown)


samuelstroschein:
Correction to the prior comment: controlled release Criterion A/B shows single-key Snapshot::get reduces missing-key latency 35.4% and 4 KiB-hit latency 30.3% versus multi_get. Each side was rebuilt in the same isolated worktree and target directory.

PR #772, opened 2026-07-25
bench(plugin-v2): profile ordinary Markdown and JSON writes

samuelstroschein:
Stacked on #754.

Adds two ignored, repeatable gates for the unchanged public SQL byte-write path:

- a steady-state 3.5 MiB Markdown profile target isolated from fixture import, Git, merges, and cold opens
- a 10 MiB real-Wasm JSON benchmark using ordinary full-byte SQL updates, asserting one semantic change and zero reparses

Baseline on this commit:
- Markdown: 42.55 / 41.88 ms mean across two sequential 100-edit runs
- JSON: 51.39 / 52.82 ms p50 across two sequential 7-edit runs

No production behavior or public API changes.


samuelstroschein:
Superseded by the consolidated Markdown/JSON benchmark gate in #783.

PR #773, opened 2026-07-25
perf(plugin-v2): vectorize fallback splice discovery

samuelstroschein:
Stacked on #772.

Profiles showed ordinary one-byte writes still perform an O(file) host fallback diff. This keeps the SQL/WIT/plugin APIs unchanged while removing the redundant equality scan and finding common prefix/suffix in machine words. Alignment coverage checks every mismatch position.

Controlled sequential results:
- Markdown 3.5 MiB: 42.21 -> 40.23 ms mean (-4.7%)
- JSON 10 MiB: 52.11 -> 44.62 ms p50 (-14.4%)

Both fixtures emit one semantic change with zero full reparses. This is a useful general improvement, but it does not yet meet the Markdown 2x-Git gate; follow-up profiling continues in the next stack layer.

Validation:
- plugin incremental unit suite: 14 passed
- release Markdown profile: 2 x 100 edits
- release 10 MiB JSON gate: 2 x 7 edits

### DIFF
--- a/packages/engine/src/plugin/incremental.rs
+++ b/packages/engine/src/plugin/incremental.rs
@@ -285,7 +285,8 @@ pub(crate) fn build_file_update_splices(
                 0,
             )
         } else {
-            if before == after {
+            let (prefix, prefix_bytes_compared) = common_prefix_len(before, after);
+            if prefix == before.len() && prefix == after.len() {
                 return Ok(BuiltInputSplices {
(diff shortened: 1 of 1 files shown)


samuelstroschein:
Superseded by #754, which now contains the vectorized fallback splice discovery together with the actor/runtime hot path.

PR #774, opened 2026-07-25
perf: add clean tracked entity read lane

samuelstroschein:
## What changed

Adds a narrow clean tracked-state serving lane for the already-native SQL
entity-primary-key read shape. When all of the following are proven from the
same read snapshot, it reads current v5 groups directly and skips the broad
untracked visibility scans:

- the request is a strict active-branch tracked entity-PK read;
- active/global heads have matching v5 markers;
- neither source has a local-sidecar marker; and
- no matching global flat sidecar row exists.

Any uncertainty returns `None` and falls back to the existing general
live-state path. Transaction readers intentionally retain that fallback so
staged overlays stay authoritative.

## Safety

Focused tests cover:

- global/local precedence and a local tombstone hiding the global row;
- a matching global mutable sidecar declining the lane; and
- a branch sidecar marker conservatively declining the lane.

The strict SQL route also has a dispatch test proving it does not re-enter the
generic scan when the capability is available.

## Before / after: release hot SQL PK read

Same host, RocksDB backend, 10k seeded tracked rows, one warmed session, and
only the operation window timed. Both use 100k repeats of the same benchmark
command; values are one-run hot-window means.

| Revision | Read one tracked PK |
| --- | ---: |
| `main` at `135a8c1a` | 30.167 us |
| This PR (`e1dc62c1`) | 28.185 us |

That is **6.6% lower latency** for this PR alone. Against the original
post-v5 pre-#771 baseline (35.704 us), the accumulated v5 + point-get + clean
lane stack is 21.1% lower. Standalone raw SQLite remains 2.696 us, so this is
still about 10.5x slower and deliberately not claimed as parity.

Reproduce:

```text
LIX_TRACKED_STATE_CRUD_PROFILE=1 \
LIX_TRACKED_STATE_CRUD_PROFILE_OP=read_one \
LIX_TRACKED_STATE_CRUD_PROFILE_LAYER=sql_session \
LIX_TRACKED_STATE_CRUD_PROFILE_HOT_REPEATS=100000 \
cargo bench -p lix_engine_benchmarks --bench tracked_state_crud --features storage-benches
```

## Validation

- `cargo test -p lix_engine clean_v5_entity_pk_lane --lib`
- `cargo test -p lix_engine native_entity_primary_key_read_uses_clean_live_state_lane_when_available --lib`
- `cargo clippy -p lix_engine --features storage-benches -- -D warnings`

### DIFF
--- a/packages/engine/src/live_state/context.rs
+++ b/packages/engine/src/live_state/context.rs
@@ -17,7 +17,7 @@ use crate::live_state::tracked_head::TrackedHeadContext;
 use crate::live_state::{
     LiveStateExactBatchRequest, LiveStateReader, LiveStateRowIdentity, LiveStateRowRequest,
     LiveStateScanRequest, MaterializedLiveStateRow, VisibilityBranchScope, VisibilityRequest,
-    expanded_branch_ids, resolve_visible_rows,
+    expanded_branch_ids, load_local_sidecar_branch_token, resolve_visible_rows,
 };
 use crate::storage_adapter::StorageAdapterRead;
 use crate::tracked_state::{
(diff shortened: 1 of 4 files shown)


samuelstroschein:
Closing by design direction: this is a narrow conditional read fast path (6.6% on one shape), not the fundamental multi-query/multi-row improvement we want. We are replacing this line with the direct tracked control + commit plane.

PR #777, opened 2026-07-25
perf(filesystem): retain materialized snapshots without reread

samuelstroschein:
## Summary

- retain the exact materialized filesystem snapshot
- avoid collecting and hashing the same 3.5 MiB file a second time after materialization
- preserve external-change detection and public APIs

## Measured result

Controlled Markdown mean improved from 40.23 ms to 39.03 ms, about 3.0%.

## Validation

- four LocalFilesystem disk synchronization regression tests
- full Git/Lix Markdown comparison in the completed stack
- 10 MiB JSON ordinary SQL gate in the completed stack

Stacked on the transaction timing PR.

### DIFF
--- a/packages/rs-sdk/src/filesystem.rs
+++ b/packages/rs-sdk/src/filesystem.rs
@@ -299,7 +299,7 @@ impl FilesystemPathFilter {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct MaterializedSnapshot {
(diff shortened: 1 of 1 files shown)


samuelstroschein:
Closing the LocalFilesystem-only retained-snapshot experiment. Its measured ~3% gain is outside the final engine + plugin + RocksDB merge stack.

PR #784, opened 2026-07-25
perf(plugin-v2): reconcile semantic writes sparsely

samuelstroschein:
## What changed

Treats direct semantic writes as same-owner plugin transitions during reconciliation, just like byte writes. Registry and file-owner resolution now use exact batched point reads instead of filtered live-state scans. Adds opt-in spans around semantic preparation, actor lookup, leasing, cold open, and rendering.

## Root cause

The reconciliation lifecycle scan only recognized selected byte-write plugins. A semantic merge therefore loaded every durable entity owned by the plugin as if it might need lifecycle replacement. On the 7,253-entity Markdown fixture this consumed roughly 100 to 200 ms even though the merge changed one entity.

## Impact

Semantic staging falls to roughly 10 ms while retaining plugin ownership and lifecycle behavior. The same fix applies to JSON and other v2 semantic plugins.

## Validation

- Mandatory engine all-simulations suite passes
- Markdown plugin suite passes
- 10 MiB JSON unrelated-property merge gate passes

### DIFF
--- a/packages/engine/src/commit_graph/walker.rs
+++ b/packages/engine/src/commit_graph/walker.rs
@@ -84,44 +84,34 @@ where
         .map(|reachable| reachable.commit.commit_id)
         .collect::<BTreeSet<_>>();
 
-    let mut best = Vec::new();
-    for reachable in left_reachable {
-        let commit_id = &reachable.commit.commit_id;
-        if !common_ids.contains(commit_id) {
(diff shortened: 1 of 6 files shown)


samuelstroschein:
This was merged only into the former intermediate benchmark branch. The final latest-main review boundary, including this work plus sparse history and linear ancestry, is #795.

PR #786, opened 2026-07-25
perf(merge): find common ancestors in one pass

samuelstroschein:
## What changed

Computes best common ancestors with one traversal per head plus one direct-parent marking pass over the common set.

## Root cause

The previous algorithm walked the reachable DAG again for every common-ancestor candidate to ask whether it had a common descendant. Repeated merges therefore grew from about 6 ms to 79 ms in merge-base discovery as shared history accumulated.

The intersection of two ancestor sets is ancestor-closed. A common commit is non-best exactly when it is a direct parent of another common commit, so no per-candidate reachability walk is needed.

## Impact

Merge-base discovery becomes linear in reachable history rather than quadratic in common history. Combined with the preceding sparse-history PR, the plain-RocksDB Markdown merge reaches p50 9.244 ms versus Git 123.851 ms.

## Validation

- All 14 commit-graph walker tests, including diamond and criss-cross histories
- cargo test -p lix_engine --features all-simulations
- 20-sample release Markdown merge benchmark


samuelstroschein:
Superseded by #795, which combines semantic reconciliation, sparse merge history, and one-pass common-ancestor discovery.

PR #789, opened 2026-07-25
perf(markdown): render semantic edits from persistent state

samuelstroschein:
## What changed

- Render Markdown semantic edits from the persistent document state.
- Preserve sparse successor state instead of rebuilding the full Markdown document for every unrelated-entity merge.
- Keep the existing SQL and Wasm Component plugin APIs unchanged.

## Why

The semantic merge itself was sparse, but materializing the resulting Markdown bytes still reconstructed more document state than the changed entities required. That made the derived-byte path dominate otherwise cheap unrelated-entity merges.

## Measured impact

Release benchmark, 3,670,017-byte Markdown corpus with 7,253 paragraphs:

- Lix unrelated-entity merge: 10.123 ms p50
- Git unrelated-line merge: 124.447 ms p50
- Lix is 12.3x faster at the median
- Lix cleanly preserves two edits to distinct cells on the same Markdown table line; Git reports a conflict

The next stacked PR further reduces protocol-digest work without changing this renderer.

## Validation

- Markdown plugin test suite
- RocksDB-backed Git-versus-Lix Markdown benchmark
- Engine all-simulations suite

### DIFF
--- a/plugins/markdown-v2/src/core.rs
+++ b/plugins/markdown-v2/src/core.rs
@@ -2027,14 +2027,14 @@ impl Document {
             .into_iter()
             .map(entity_change_to_detected)
             .collect::<Result<Vec<_>, _>>()?;
-        if let Some((bytes, top_level_ranges, edits, root)) =
+        if let Some((bytes, top_level_ranges, edits, tree)) =
             self.try_paragraph_entity_change(&detected)?
         {
             return Ok((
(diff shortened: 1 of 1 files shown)


samuelstroschein:
Superseded by #779, which now combines persistent Markdown successors with sparse semantic rendering.

PR #827, opened 2026-07-27
Add change proposal API

samuelstroschein:
## Summary

- add tracked global `lix_change_proposal` records with create, inspect, accept, and reject APIs
- add `lix_branch_diff(source_branch_id, target_branch_id)` and conflict review surfaces, using `diff(merge_base(source, target), source)`
- preserve proposal snapshot heads for stable review/acceptance, guard source and target heads atomically, and retain open snapshots during GC
- expose the API through the Rust SDK and document the model, trade-offs, and benchmark results

## Design

A proposal is a first-class tracked global entity rather than key-value metadata. It deliberately has no duplicated lifecycle timestamps: Lix change metadata provides creation/update provenance. Acceptance uses the pinned review snapshot and applies the merge plus proposal transition as one durable transaction; reject only changes proposal state.

## Validation

- `git diff --check origin/main...HEAD`
- `cargo check -j 1 -p lix_engine --test change_proposal`
- `cargo check -j 1 -p lix_sdk --no-default-features`
- focused proposal test suite previously passed: 14 tests


samuelstroschein:
Closing this long-stale, conflicting draft during the v0.14.0 PR cleanup. Its branch and discussion remain available if the proposal is revived against current main.

PR #877, opened 2026-07-28
perf: certify normalized replacement batches

samuelstroschein:
## Summary

Stacked on #876.

- derive a narrow certificate for complete `{ path, value }` schemas whose `value` accepts every JSON type and whose identity is the already-valid string `path`
- lower homogeneous bound `UPDATE ... SET value = lix_json($n) WHERE path = $m` batches directly into canonical replacement bytes
- project only identity/system fields plus metadata from live state, avoiding predecessor snapshot hydration and full replacement-object construction
- keep decoded JSON lazy after the certified normalization boundary
- retain the canonical path for inserts, untracked rows, file-owned rows, checkpoints/lifecycle work, constrained schemas, repeated identities, and non-SQL callers

No public SQL API or storage-adapter API changes.

## Profile evidence

10,000 distinct bound updates, RocksDB, 15 order-balanced parent/candidate pairs, each process reporting the median of 5 independently seeded fixtures:

| build | median |
| --- | ---: |
| #876 parent | 134.803 ms |
| candidate | 110.907 ms |

- independent median: **17.73% faster**
- paired median: **17.89% faster**
- candidate wins: **14/15**
- raw SQLite median: 13.289 ms
- current Lix/SQLite ratio: about **8.35x** (still above the 2x project goal)

The profile attributed the removable ownership cluster to predecessor JSON deserialization, BTree object construction/cloning/insertion/drop, normalization, and allocation. Earlier storage experiments did not clear 10%, so this PR does not widen the storage adapter.

## Regression controls

| workload | parent | candidate | delta |
| --- | ---: | ---: | ---: |
| transaction insert 10k | 63.363 ms | 64.587 ms | +1.93% |
| transaction update 10k | 79.611 ms | 79.699 ms | +0.11% |
| transaction delete 10k | 79.632 ms | 79.614 ms | -0.02% |
| populated working diff | 0.531 ms | 0.517 ms | -2.64% |
| merge preview | 0.134 ms | 0.139 ms | +3.73% |
| merge commit | 0.518 ms | 0.516 ms | -0.39% |

Diff and merge figures are medians from 9 order-balanced pairs. CRUD controls use 9 pairs except insert, which uses 15.

## Rejected alternative

Materializing the complete replacement DOM at the prepared-row boundary restored the old shared layout but measured about 137 ms for the target workload and did not clear the 10% gate. The accepted representation restores the original `TransactionJson` size and keeps laziness only on `StageJson`, where mixed-transaction validation can materialize certified content if required.

## Correctness

- schema proof accepts both JSON Schema `anyOf` and `type: [...]` forms for every JSON value type
- schemas with value constraints or unsupported top-level constraints remain ineligible
- certified execution test proves the fast lane ran and round-trips JSON null plus nested JSON
- prepared-boundary test proves certified content remains materializable for mixed validation

## Validation

  - 1,397 unit tests
  - 760 integration simulations (2 ignored)
  - 3 doctests

### DIFF
--- a/packages/engine/src/session/execute.rs
+++ b/packages/engine/src/session/execute.rs
@@ -3689,6 +3689,87 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn execute_batch_certifies_complete_path_value_replacements() {
+        let session = open_session().await;
+        let schema = serde_json::json!({
+            "x-lix-key": "certified_replacement_probe",
+            "x-lix-primary-key": ["/path"],
+            "type": "object",
+            "properties": {
+                "path": { "type": "string" },
+                "value": {
+                    "type": [
+                        "object", "array", "string", "number", "integer", "boolean", "null"
+                    ]
+                }
+            },
+            "required": ["path", "value"],
+            "additionalProperties": false
+        });
+        session
+            .execute(
+                "INSERT INTO lix_registered_schema (value) VALUES (lix_json($1))",
+                &[Value::Text(schema.to_string())],
+            )
+            .await
+            .unwrap();
+        session
+            .execute(
+                "INSERT INTO certified_replacement_probe (path, value) VALUES \
+                 ('/a', lix_json('\"old-a\"')), ('/b', lix_json('\"old-b\"'))",
+                &[],
+            )
+            .await
+            .unwrap();
+
+        sql2::take_certified_replacement_parameter_batch_executions();
+        let sql = "UPDATE certified_replacement_probe SET value = lix_json($1) WHERE path = $2";
+        let results = session
+            .execute_batch(&[
+                ExecuteBatchStatement {
+                    sql: sql.to_string(),
+                    params: vec![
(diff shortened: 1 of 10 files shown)


samuelstroschein:
Post-#877 profile and rejected follow-up:

- CPU moved from predecessor/full-row JSON ownership into RocksDB encoding and comparisons: `memcmp` 11.0%, `memmove` 5.7%, CRC 3.5%, BLAKE3 2.5%, Zstd/Huffman/FSE about 5%+, RocksDB comparators about 4%.
- Upper-bound experiment: disable SST compression entirely.
- 11 order-balanced pairs, 5 fixtures per process: Zstd median **113.21 ms**, no-compression median **118.38 ms**.
- No compression is **4.6% slower**, so the storage change was fully reverted and no storage-adapter API was added or removed.

The next cut should not infer that compression CPU is removable; Zstd is reducing enough copying/write amplification to be net-positive.

PR #878, opened 2026-07-28
perf(engine): accelerate root file listings

samuelstroschein:
## Summary

- recognize the exact no-plugin root `lix_file` listing shape and return it directly from the shared filesystem path index, skipping DataFusion catalog/planning/Arrow work
- preserve all other SQL shapes on the existing relational path with a deliberately strict recognizer
- fix a SlateDB snapshot/publication race found while profiling the real file-write benchmark
- extend the public API benchmark to measure both the file query and the complete directory-plus-file listing on Memory, RocksDB, and SlateDB

## Why

The warmed root listing spent most of its time allocating and planning two DataFusion queries even though the shared filesystem path index already held every projected value. The direct route benefits every adapter rather than specializing for SlateDB.

The write benchmark also exposed a correctness race: a read could fetch a stale SlateDB snapshot immediately before a publication persisted, while an unrelated view drop reclaimed that publication before the stale fetch captured it. The fix retains publications through the newest actually observed snapshot and refuses to cache snapshots that do not cover all persisted publications.

## Performance

Measured from separate release binaries against latest `origin/main` `564a5c8d9`, 100-file corpus, 30 warmups, 300 samples per backend. Values below are the average p50 from two process runs.

| Public API operation | Backend | main | PR | Improvement |
|---|---:|---:|---:|---:|
| complete root listing (directories + files) | RocksDB | 1.240 ms | 603 µs | 51.4% faster |
| complete root listing (directories + files) | SlateDB | 1.215 ms | 592 µs | 51.3% faster |
| root file query | RocksDB | 647 µs | 27.0 µs | 95.8% faster |
| root file query | SlateDB | 656 µs | 24.9 µs | 96.2% faster |

The previously conflicting SlateDB single-overwrite and 10-file batch-overwrite Criterion workloads both completed warmup and sampling without a transaction conflict.

## Correctness and validation

- strict route/fallback recognition tests
- direct-result equivalence test covering root/nested files, order, JSON metadata, null metadata, and timestamps
- deterministic publication reclamation race regression test
- stale snapshot cache regression test
- `cargo test --profile test --workspace --all-features --no-run`

No storage contract change is required for this cut.

### DIFF
--- a/packages/engine-benchmarks/tests/exact_file_read_benchmark.rs
+++ b/packages/engine-benchmarks/tests/exact_file_read_benchmark.rs
@@ -213,6 +213,74 @@ where
             percentile(&samples, 95).as_nanos(),
         );
     }
+
+    let root_listing_sql = "SELECT id, path, name, lixcol_metadata, lixcol_updated_at \
+         FROM lix_file WHERE directory_id IS NULL ORDER BY name";
+    for _ in 0..WARMUPS {
+        black_box(
+            session
+                .execute(root_listing_sql, &[])
+                .await
+                .expect("warm root listing"),
+        );
+    }
+    let mut samples = Vec::with_capacity(ROUNDS);
+    for _ in 0..ROUNDS {
+        let started = Instant::now();
+        let result = session
+            .execute(root_listing_sql, &[])
+            .await
+            .expect("execute root listing");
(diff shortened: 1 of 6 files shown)


samuelstroschein:
@codex review

PR #879, opened 2026-07-28
perf(engine): accelerate root directory listings

samuelstroschein:
## Stack

- Base: #878 (`30c37d15b`), green and reviewed
- This PR contains only the second root-directory cut

## Summary

- hard-cut the internal file-only route name into `ExactFilesystemRead`
- recognize the exact root `lix_directory` listing used by the no-plugin filesystem API
- return directory rows directly from the shared filesystem path index, preserving every other SQL shape on DataFusion
- extend strict recognition and relational-equivalence coverage to root directories

## Performance versus the previous stack baseline

Separate release binaries, 100-file corpus, 30 warmups, 300 samples per backend, average p50 from two process runs:

| Public API operation | Backend | #878 baseline | This PR | Improvement |
|---|---:|---:|---:|---:|
| complete root listing (directories + files) | RocksDB | 603 µs | 39.9 µs | 93.4% faster |
| complete root listing (directories + files) | SlateDB | 592 µs | 37.0 µs | 93.7% faster |
| already-fast root file query | RocksDB | 27.0 µs | 26.7 µs | flat / no regression |
| already-fast root file query | SlateDB | 24.9 µs | 24.9 µs | flat / no regression |

This is a shared engine cut: both RocksDB and SlateDB improve by essentially the same amount.

## Correctness and validation

- strict route/fallback tests reject aliases, alternate ordering, descending order, and limits
- direct directory output equals the relational path for ordering, nested exclusion, ids, paths, names, and timestamps
- `cargo test --profile test -p lix_engine --all-features`

No storage contract change is required.


samuelstroschein:
@codex review

PR #880, opened 2026-07-28
perf: load minimal replacement owners

samuelstroschein:
## Summary

- hard-cut certified replacement candidate hydration to an aligned minimal ownership projection
- preserve missing-row affected counts and inherited metadata without materializing full live-state rows
- decline staged, global-schema, file-scoped, untracked, and unsupported visibility cases to the canonical path
- leave the storage adapter API unchanged because the profiled point-probe-to-scan experiment did not clear the acceptance bar

## Profile and rejected cut

After #877, `memcmp` was 11.02% of the target profile and about 76% of sampled `memcmp` calls came from RocksDB comparators. Replacing 10,000 point probes with an ordered range scan was therefore tested first. Eleven balanced hot pairs improved only about 2.4%, while `memcmp` remained 10.63%; the iterator performs substantially the same comparator work and adds scan/page overhead. That experiment was reverted.

`memmove` attribution then showed repeated `String` cloning and construction of generic live-state `BTreeMap`s. Certified replacements only need aligned existence plus ownership metadata. Bypassing full generic candidate hydration established an approximately 28% ceiling, so this change introduces the smallest correct projection for that evidence.

## Benchmarks

All target measurements use balanced, fresh-fixture pairs.

| Benchmark | Parent | Candidate | Change |
| --- | ---: | ---: | ---: |
| 10k certified bound UPDATE | 117.71 ms | 98.21 ms | **16.6% faster** |
| Raw SQLite reference | 13.289 ms | — | candidate is 7.39× SQLite |
| SQL-session INSERT | 138.659 ms | 139.814 ms | 0.8% slower |
| Ordinary SQL UPDATE | 406.539 ms | 425.653 ms | 4.7% slower |
| DELETE | 110.735 ms | 114.197 ms | 3.1% slower |
| Transaction UPDATE | 84.131 ms | 84.720 ms | 0.7% slower |
| Diff | 30.128 ms | 30.503 ms | 1.2% slower |
| Merge preview | 121.333 ms | 123.600 ms | 1.9% slower |
| Merge commit | 16.571 ms | 13.469 ms | 18.7% faster |

The target won all 15/15 balanced pairs. In the candidate profile, `memcmp` falls to 7.34%, and the generic visibility `BTreeMap` disappears from the top costs.

## Correctness contract

- owner slots stay aligned with requested identities
- missing rows remain missing and report zero affected rows
- inherited metadata is preserved
- staged rows and visibility shapes not proven equivalent continue through the canonical reader
- checkpoint and lifecycle paths are unchanged

## Validation

- focused certified replacement end-to-end test, including missing rows and metadata preservation
  - 1,391 unit tests passed, 6 ignored
  - 760 integration tests passed, 2 ignored
  - 3 doctests passed

### DIFF
--- a/packages/engine/src/live_state/context.rs
+++ b/packages/engine/src/live_state/context.rs
@@ -12,9 +12,9 @@ use crate::filesystem::{
 };
 use crate::live_state::tracked_head::TrackedHeadContext;
 use crate::live_state::{
-    LiveStateExactBatchRequest, LiveStateReader, LiveStateRowFilter, LiveStateRowIdentity,
-    LiveStateRowRequest, LiveStateScanRequest, MaterializedLiveStateRow, VisibilityBranchScope,
-    VisibilityRequest, expanded_branch_ids, resolve_visible_rows,
+    LiveStateExactBatchRequest, LiveStateReader, LiveStateReplacementOwner, LiveStateRowFilter,
+    LiveStateRowIdentity, LiveStateRowRequest, LiveStateScanRequest, MaterializedLiveStateRow,
+    VisibilityBranchScope, VisibilityRequest, expanded_branch_ids, resolve_visible_rows,
 };
 use crate::storage_adapter::StorageAdapterRead;
 use crate::tracked_state::{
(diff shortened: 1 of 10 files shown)


samuelstroschein:
Closing after rebasing and re-measuring against current main (45d154e05, UUIDv7 physical keys). In 15 balanced fresh-fixture pairs, main median was 99.79 ms and the rebased candidate median was 102.77 ms: the candidate is 3.0% slower. The physical-key merge removed the generic hydration cost this cut targeted, so it no longer clears the >10% acceptance bar and should not land.

PR #881, opened 2026-07-28
perf(slatedb): advance point cache across commits

samuelstroschein:
## Summary

- carry unchanged snapshot point-cache entries across exactly contiguous local SlateDB commits
- replace or invalidate every key written by the successful commit before exposing the advanced cache bucket
- clear the cache on external or otherwise non-contiguous sequence changes
- repair the exact-file benchmark fixture for UUID file IDs and enable its required multithread Tokio runtime

The storage contract and RocksDB implementation are unchanged. Active older snapshots remain correct: a moved cache entry is only an optimization, and an older reader refetches on a miss.

## Performance

Native file upsert component benchmark, 5,000 files + 5,000 history rows, 30 paired accepted-commit samples on the same host:

| Adapter / revision | Sequential CRUD mean | Batched CRUD mean |
| --- | ---: | ---: |
| SlateDB `origin/main` (`b6eacab77`) | 5.690 ms | 2.436 ms |
| SlateDB this PR (repeat) | 4.831 ms | 2.148 ms |
| Improvement | **15.1%** | **11.8%** |
| RocksDB control (unchanged source) | 3.889 ms | 2.053 ms |

SlateDB is 1.24x RocksDB in the repeated sequential run. A separate candidate run measured 4.728 ms (1.22x RocksDB).

## Correctness and validation

- `cargo test --profile test -p lix_slatedb_storage --all-features` (31 unit + 8 conformance tests)
- full engine suite before final cleanup: 1,403 unit tests and 764 integration tests green
- `cargo clippy -p lix_slatedb_storage --all-targets --all-features -- -D warnings`
- `cargo test --profile test -p lix_engine_benchmarks --test exact_file_read_benchmark --no-run`
- `cargo fmt --all`
- `git diff --check`

## Cache safety invariant

A cache bucket advances only from sequence N to N+1 after the local database write succeeds. All committed overlay keys replace their prior cached values. Any unknown sequence, including an external writer jump, clears unchanged cached values instead of carrying them forward.

### DIFF
--- a/packages/engine-benchmarks/Cargo.toml
+++ b/packages/engine-benchmarks/Cargo.toml
@@ -42,7 +42,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 smallvec = "1.15"
 tempfile = "3"
-tokio = { version = "1", features = ["rt", "macros", "sync", "time"] }
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "time"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 xxhash-rust = { version = "0.8", features = ["xxh3"] }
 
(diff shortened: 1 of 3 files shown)


samuelstroschein:
@codex review

PR #882, opened 2026-07-28
feat(plugin): infer keyless creates from UUID defaults

samuelstroschein:
## Summary

Hard-cut the plugin create protocol from `x-lix-id-allocation` and plugin-owned ID namespaces to schema-inferred keyless creates.

- A schema is creatable when every primary-key field can be completed by a host-evaluated default. The shipped schemas use `x-lix-default: "lix_uuid_v7()"`.
- Plugins send `Create { schema_key, local_ref, snapshot }` without an ID. The host materializes the UUID, inserts it into the snapshot, derives the typed primary key, validates, and stages a normal upsert.
- Exact `lix_uuid_v7()` and `lix_timestamp()` defaults compile to direct intrinsic fast paths; other expressions continue through CEL.
- Retry-stable UUID allocation is derived from one durable create-context reservation plus the page-local `local_ref`; there is no per-entity reservation row.
- The WIT contract, public plugin API, built-in plugins, typed SQL writes, and merge projection were migrated together.

## Deliberate hard cut

There is no compatibility shim:

- `x-lix-id-allocation` and the old ID-namespace protocol are removed.
- The WIT input changes from `ids` to `creates`, and the packet protocol gains an explicit keyless `Create` mutation.
- Plugin registry format is bumped from v3 to v4, so old plugin archives are rejected instead of being silently reinterpreted.

## Performance evidence

Release RocksDB import benchmark, 10.68 MB CSV / 220,001 rows / five alternating pairs:

| Path | p50 |
|---|---:|
| Keyless plugin create | 4,488.874 ms |
| Direct typed SQL control | 4,954.526 ms |

Aggregate ratio: `0.906`; paired p50 ratio: `0.904`. In this run the keyless plugin path was about 9–10% faster than the direct semantic-row control. Packet decode was 91–114 ms and change-row materialization was 104–150 ms, confirming the create path does not pay per-row general CEL evaluation.

The focused 220k-row default benchmark measured direct UUID generation/insertion at ~27 ms versus ~398 ms through general CEL (~14.6× faster), which is why the two exact expressions have compiled intrinsic paths.

## Validation

- `cargo test --profile test --workspace --all-features --no-run`
- SlateDB unit and storage-conformance suites (35 tests)
- `cargo test --release -p lix_sdk_tests --test e2e v2_csv_ten_mib_rocksdb_import_parity_benchmark -- --ignored --exact --nocapture`
- WIT copies verified byte-identical; `git diff --check` clean


samuelstroschein:
@codex review

samuelstroschein:
@codex review

samuelstroschein:
@codex review

PR #901, opened 2026-07-28
Scale plugin imports by live working set, not document count

samuelstroschein:
## Why

An atomic plugin-backed import retained one Wasm Store per completed document transition until commit. The default 16-Store admission ceiling therefore rejected the 17th file (and the reported 1,184-file Markdown import), even though completed publications no longer needed a warm Store for transaction correctness.

## What changed

- Treat the Store ceiling as a live working-set bound rather than a document-count bound.
- Under admission pressure, retire the oldest completed **fresh- or existing-document** Store after its durable rows/materialization are staged, then reuse that permit inside the same atomic transaction.
- Preserve actively contested same-file leases: cache eviction still refuses slots referenced by an active or waiting transition.
- Publish recycled transitions as non-authoritative file views so later edits cold-open from durable state.
- Recover retained session observations from benign cache eviction only while their exact durable semantic root is unchanged; genuinely newer roots remain stale.
- Remove the hidden `execute_batch_for_single_writer_ingest(..., retain_plugin_actors)` compatibility path and its policy boolean.
- Move Git replay to the general public `execute_batch` API.
- Add production-Wasm conformance coverage for Markdown (1,184 files), CSV, text, and Excalidraw, including committed bytes and semantic-row verification.
- Add a deterministic two-batch Git-text fixture that imports 17 files and then updates them in path order, forcing observation-named Store eviction between batches.
- Expand Git replay coverage to a 100-commit repository whose second commit atomically updates 17 existing plugin-owned text files.
- Check in the benchmark/profiling procedure and results.

## Performance

Release profile, Linux x86-64, Rust nightly 1.97.0, seven fresh warm-cache processes, one atomic 1,184-document Markdown statement (30,932 logical input bytes):

| Metric | Result |
| --- | ---: |
| p50 | 594.637 ms |
| p95 | 604.307 ms |
| p50 / document | 0.502 ms |
| whole-process wall | 680 ms |
| peak RSS | 148,848 KiB |
| swaps | 0 |

A profiled run completed the measured import in 577.842 ms with 775.27 ms task-clock, 4.58B instructions, and 41.6M cache misses.

Exact pre-change `origin/main` (`f860b30ea`) fails the capacity-plus-one Markdown fixture with `LIX_ERROR_PLUGIN_RESOURCE_LIMIT`; this branch keeps the same 16-Store resource ceiling and passes 1,184 fresh files plus broad existing-file updates atomically.

## Validation

- `cargo clippy --profile test -p lix_engine -p lix_sdk -p lix_cli -p lix_sdk_tests --all-targets --all-features -- -D warnings`
- `cargo test -p lix_sdk_tests --test plugin_document_scale -- --nocapture --test-threads=1`
- `cargo test -p lix_cli git_replay`
- release fixture: 7-run distribution, `/usr/bin/time`, and `perf stat`

This intentionally removes the unreleased specialized ingestion API; no backward-compatibility shim is retained.

### DIFF
--- a/.changenotes/scale-plugin-imports-by-working-set.md
+++ b/.changenotes/scale-plugin-imports-by-working-set.md
@@ -1,0 +1,8 @@
+---
+type: minor
+---
+
+Plugin-backed atomic imports now scale independently of document count. The
+engine automatically reuses its bounded live-Store working set for fresh
+documents while retaining existing-document leases through commit, so callers
+no longer need a special single-writer ingestion API or actor-retention policy.
--- a/.changenotes/scale-plugin-imports-by-working-set.md
+++ b/.changenotes/scale-plugin-imports-by-working-set.md
@@ -3,6 +3,7 @@ type: minor
 ---
 
 Plugin-backed atomic imports now scale independently of document count. The
-engine automatically reuses its bounded live-Store working set for fresh
-documents while retaining existing-document leases through commit, so callers
-no longer need a special single-writer ingestion API or actor-retention policy.
+engine automatically reuses its bounded live-Store working set for fresh and
+existing documents while preserving actively contested same-file leases, so
+callers no longer need a special single-writer ingestion API or
+actor-retention policy.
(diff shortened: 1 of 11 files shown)


samuelstroschein:
@codex review

PR #985, opened 2026-07-30
compress packed commit-delta sidecars

samuelstroschein:
## What changed

- Hard-cut the packed commit-delta format from `LXCD5` to `LXCD6` and bump the repository protocol to V24.
- Keep the front-coded identity/index leaf uncompressed, so diff/checkpoint/index-only readers retain their allocation-free fast path.
- Compress only the eagerly persisted payload sidecar with zstd level 1, with an explicit codec byte, bounded decoded length, corruption validation, and raw fallback when compression would expand data.
- Decode compressed sidecars into owned storage only for payload-consuming history/merge reads.

## Data-driven result

Retained SlateDB replay: VS Code Docs, exact same 50 commits, Markdown/text/CSV plugins enabled, eager materialization, checkpoint and tree verification after every Git commit.

- Total live logical storage: **176.56 MB → 147.88 MB (-16.2%)**
- Commit-delta segment storage: **52.97 MB → 24.51 MB (-53.7%)**
- Payload sidecar: **40.29 MB → 11.87 MB (-70.5%)**
- Rows retained: 302,366 commit-delta rows; final tree verified.

The retained candidate store is at `/root/projects/openclaw-perf-data/vscode-sidecar-zstd-50-db` and its accounting is at `/root/projects/openclaw-perf-data/vscode-sidecar-zstd-50-layout.txt`.

The first end-to-end replay sample was +2.5% versus the preceding run; that delta was entirely concentrated in the same two very large commits while the median commit was 2.5 ms faster. A controlled 100k-history raw/compressed A/B measured raw seeding at 1.731–1.739 s and compressed at 1.723–1.866 s (median +1.3%) while logical bytes written fell 3.81 MB → 3.21 MB (-15.9%). This does not justify adding thread-pool overhead or WASM complexity to the codec path.

Physical SlateDB directory size is intentionally not used as the live-data metric here because old SST/WAL objects remain until compaction; the exact per-space live logical inventory above is read from the retained stores.

## Validation

- `cargo test -p lix_engine --features all-simulations` (1,590 unit tests, 782 simulations, 3 doctests)
- focused packed commit-delta, indexed payload, corruption, protocol-gate, and locator tests
- `cargo clippy -p lix_engine --all-targets --features all-simulations` (green; existing main dead-code warnings remain)
- `cargo check -p lix_js_sdk --target wasm32-unknown-unknown`
- replay verified every commit and the final 8k-file tree

### DIFF
--- a/.changenotes/compress-commit-delta-sidecars.md
+++ b/.changenotes/compress-commit-delta-sidecars.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+Packed tracked-history sidecars now use bounded zstd compression while identity-only reads retain their uncompressed fast path.
--- a/packages/engine/src/engine.rs
+++ b/packages/engine/src/engine.rs
@@ -666,6 +666,30 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn predecessor_v23_commit_delta_protocol_is_rejected() {
+        let storage = Memory::new();
+        Engine::initialize(storage.clone())
+            .await
+            .expect("engine should initialize");
+        let storage_adapter = StorageAdapter::new(storage.clone());
+        let mut writes = storage_adapter.new_write_set();
+        writes.put(
+            crate::init::REPOSITORY_PROTOCOL_SPACE,
+            crate::init::REPOSITORY_PROTOCOL_KEY,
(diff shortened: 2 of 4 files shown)


samuelstroschein:
@codex review

PR #988, opened 2026-07-30
inline medium JSON without diff regressions

samuelstroschein:
## What changed

- raise the deterministic JSON inline threshold from 256 to 1,024 bytes
- add a V6 HOT row slot that persists content fingerprints only while inline rows are dirty
- make identity-only SQL diff surfaces validate allocation-free commit-delta leaves and hydrate payloads only for live/live equality checks
- retain full eager payload batches for merge and checkpoint consumers
- bump the repository protocol to V25; no migration or backward compatibility

## Why

Retained real-repository profiling showed that medium semantic JSON dominated `json_store` indirection. Raising the threshold removes those rows, but a constant-only change made working diff repeatedly hash inline payloads and made historical SQL diff decompress payload sidecars for rows whose classification did not require content. Dirty-only fingerprints and projection-aware eager hydration remove both regressions.

## Retained replay

VS Code Docs, 50 first-parent commits, all bundled WASM plugins, eager semantic materialization, SlateDB, checkpoint every Git commit, changed-file verification after every commit, and final Git tree verification:

- logical bytes: 147,883,439 -> 117,975,640 (-20.2%)
- physical bytes: 105,435,166 -> 89,260,884 (-15.3%)
- replay: 115,432.867 ms -> 115,383.104 ms (-0.04%)
- execute: 97,075.792 ms -> 97,301.970 ms (+0.23%)
- checkpoint: 4,315.573 ms -> 4,154.759 ms (-3.7%)
- final tree verification: 12,619.835 ms -> 12,551.529 ms (-0.5%)

The retained candidate database is:

`/root/projects/openclaw-perf-data/vscode-json-inline-fingerprint-50-db`

## CRUD/diff A/B

Memory storage, 1,000 rows with 800-byte values, seven warm samples, compared with the preserved 256-byte baseline:

- working diff: 2.291 -> 2.236 ms (-2.4%)
- revert 100: 5.698 -> 5.502 ms (-3.4%)
- apply 100: 6.749 -> 6.031 ms (-10.6%)
- partial checkpoint 500: 18.368 -> 17.066 ms (-7.1%)

## Validation

  - 1,593 unit tests passed
  - 782 simulation/integration tests passed, 2 ignored
  - 3 doc tests passed
- focused HOT codec, working-diff, protocol rejection, corruption, and tracked-state rebuild regressions
- retained replay verified all 50 per-commit states and the complete final Git tree

### DIFF
--- a/packages/engine/src/engine.rs
+++ b/packages/engine/src/engine.rs
@@ -690,6 +690,30 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn predecessor_v24_hot_inline_protocol_is_rejected() {
+        let storage = Memory::new();
+        Engine::initialize(storage.clone())
+            .await
+            .expect("engine should initialize");
+        let storage_adapter = StorageAdapter::new(storage.clone());
+        let mut writes = storage_adapter.new_write_set();
+        writes.put(
+            crate::init::REPOSITORY_PROTOCOL_SPACE,
+            crate::init::REPOSITORY_PROTOCOL_KEY,
+            &b"commit-delta-sidecar-zstd.v24"[..],
+        );
+        storage_adapter
+            .commit_write_set(writes, StorageWriteOptions::default())
+            .await
+            .expect("V24 protocol marker should commit");
+
+        let Err(error) = Engine::new(storage).await else {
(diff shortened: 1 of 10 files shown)


samuelstroschein:
@codex review

PR #990, opened 2026-07-30
fall back on structural plugin edits

samuelstroschein:
## What changed

- Treat JSON scalar parse failures in the sparse arena path as optimization misses, allowing full-document reconciliation.
- Treat Excalidraw element parse failures in the sparse span path the same way.
- Preserve validation errors from successfully parsed scalar/element records.
- Add unit and real Wasm end-to-end regressions for sibling/element insertion at an indexed span boundary.

## Why

The sparse paths assumed an edit contained by an existing scalar or element span must still describe exactly one scalar or element. Valid structural inserts at the old span boundary violate that assumption: the local slice contains adjacent syntax while the complete successor remains valid. Rejecting the local parse prevented the intended full-document fallback.

## Validation

- `cargo test -p plugin_json`
- `cargo test -p plugin_excalidraw`
- `cargo test -p lix_sdk_tests --test e2e json_scalar_boundary_insert_adds_sibling_through_full_reconciliation -- --nocapture`
- `cargo test -p lix_sdk_tests --test e2e excalidraw_element_boundary_insert_adds_element_through_full_reconciliation -- --nocapture`
- `git diff --check`

### DIFF
--- a/packages/rs-sdk-tests/tests/e2e.rs
+++ b/packages/rs-sdk-tests/tests/e2e.rs
@@ -2502,6 +2502,42 @@ async fn json_scalar_to_container_edit_uses_full_reconciliation() {
 }
 
 #[tokio::test]
+async fn json_scalar_boundary_insert_adds_sibling_through_full_reconciliation() {
+    let lix = open_lix(OpenLixOptions::default()).await.unwrap();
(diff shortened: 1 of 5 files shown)


samuelstroschein:
@codex review

PR #991, opened 2026-07-30
Fix Markdown arena EOF ranges

samuelstroschein:
## What changed

- clamp serialized Markdown arena block ranges to the exact accepted byte length
- include block ordinal and exact range in future invalid-state diagnostics
- add a regression proving parser-only EOF sentinels are never persisted

## Root cause

The Markdown parser can expose top-level ranges ending at a virtual two-byte EOF sentinel. `arena_state()` persisted those parser coordinates directly. After actor eviction/hydration, the next transition rejected the durable arena because a block ended at byte 4002 for an exact 4000-byte accepted file.

## Validation

- `cargo test -p plugin_markdown persisted_arena_ranges_exclude_parser_only_eof_sentinels -- --nocapture`
- exact six-commit VS Code docs SlateDB replay with all plugins, checkpoints, and `--verify-state`
- replay completed 6/6 commits and verified the final Git tree manifest

### DIFF
--- a/plugins/markdown/src/core.rs
+++ b/plugins/markdown/src/core.rs
@@ -2115,7 +2115,7 @@ impl Document {
         }
         let mut children = Vec::with_capacity(blocks.len());
         let mut ranges = Vec::with_capacity(blocks.len());
-        for block in blocks {
+        for (ordinal, block) in blocks.into_iter().enumerate() {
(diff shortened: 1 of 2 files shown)


samuelstroschein:
@codex review

PR #999, opened 2026-07-30
bound lazy snapshot reads by transition page limit

samuelstroschein:
## What changed

- cap plugin SDK lazy entity-snapshot reads by the transition-advertised `max_page_bytes`
- preserve the existing 1 MiB preferred chunk ceiling when the transition permits it
- extend the direct JSON cold-successor regression with a 96 KiB lazy durable snapshot under a 32 KiB page limit

## Why

Plugin API v4 correctly pages entity records at the advertised bound, but lazy snapshot attachments still requested fixed 1 MiB chunks. Hosts reject those imports with `LimitExceeded` whenever a transition uses a smaller page limit. The shared `EntityChangeReader` serves hydration, semantic changes, and cold successors, so bounding its attachment reads fixes all three paths.

This is the isolated follow-up to the exact-head review finding intentionally deferred while merging #995 to unblock retained replay work.

## Validation

- `cargo test -p lix_sdk_tests --test e2e v3_json_direct_cold_successor_preserves_durable_identity -- --exact --nocapture`
- `cargo clippy -p lix_plugin_api -p lix_sdk_tests --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/packages/plugin-api/src/lib.rs
+++ b/packages/plugin-api/src/lib.rs
@@ -539,7 +539,12 @@ impl EntityChangeReader<'_> {
                 }
                 Some(snapshot)
             }
-            (Some(length), None) => Some(read_entity_snapshot(self.source, index, length)?),
+            (Some(length), None) => Some(read_entity_snapshot(
+                self.source,
+                index,
+                length,
+                self.max_page_bytes,
+            )?),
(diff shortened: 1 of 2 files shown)


samuelstroschein:
@codex review

PR #1005, opened 2026-07-30
remove false plugin benchmark baselines

samuelstroschein:
## What changed

- remove duplicate v2/v3 lanes that both built the current v3 plugin archive
- keep one honest current-v3 lane for Markdown, JSON, CSV, and Excalidraw benchmarks
- retain v2 comparisons as cross-revision benchmark evidence instead of reintroducing the deleted v2 runtime
- rename the remaining generic candidate gate so it does not imply a v2/v3 comparison

## Root cause

The API hard cut removed the v2 runtime and plugin builders, but several ignored benchmarks were mechanically changed to build the v3 archive for both lane labels. Their timings and allocation counters therefore compared v3 against itself and could fail from lane ordering or noise.

## Validation

- cargo fmt --check
- cargo test -p lix_sdk_tests --test e2e --no-run (native target linked with mold)
- v3_json_ten_mib_push_sink_benchmark, 2 samples
- v3_file_changed_push_sink_benchmark, 2 samples
- v3_markdown_vscode_api_exact_transition_benchmark, 2 samples
- v3_excalidraw_large_transition_benchmark, 2 samples

Each executed lane passed exact semantic/output checks and reported one guest export.

### DIFF
--- a/packages/rs-sdk-tests/tests/e2e.rs
+++ b/packages/rs-sdk-tests/tests/e2e.rs
@@ -1198,7 +1198,7 @@ async fn v3_markdown_noncanonical_source_stays_in_file_arena_not_semantic_root()
 }
 
 #[tokio::test]
-#[ignore = "exact VS Code Docs d5badf Markdown transition v2 versus v3 benchmark"]
+#[ignore = "exact VS Code Docs d5badf Markdown transition benchmark"]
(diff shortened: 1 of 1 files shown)


samuelstroschein:
@codex review

PR #1007, opened 2026-07-30
perf(engine): retain unchanged HOT rows across checkpoints

samuelstroschein:
## What changed

- add checkpoint ownership to dirty HOT baselines and hard-cut the row codec/repository protocol to V27
- retain checkpoint-selected HOT rows when their immutable change, deletion state, and timestamps already match
- continue publishing collection-generation markers because their commit ownership is a visibility fence
- write collection controls only when their generation or live count changes
- add a SlateDB SST-space profiler for physical attribution

## Why

Checkpoint publication was rewriting nearly the complete eagerly materialized plugin state only to clear row-local working-diff baselines. On the retained 50-commit `vscode-docs` replay, HOT rows were 1.962 physical versions per key and collection controls were 1.982 versions per key. The initial checkpoint alone duplicated about 53.8k HOT rows and 30.6k controls.

Dirty rows now carry the checkpoint that owns their before-image. Advancing the working-diff epoch makes an old baseline stale without rewriting an otherwise identical current-state row. Squashed rows and collection-generation fences still publish normally.

## Measured impact

All measurements use SlateDB, all bundled WASM plugins, eager semantic materialization, verification after every commit, and one Lix checkpoint per Git commit.

Retained 50-commit `vscode-docs` replay:

- end-to-end database bytes: 50,348,872 -> 43,365,071 (-13.9%)
- compacted SST bytes: 49,722,940 -> 43,301,163 (-12.9%)
- HOT compressed bytes: 12,956,273 -> 8,260,546 (-36.2%)
- HOT versions/key: 1.962 -> 1.014
- collection-control compressed bytes: 1,605,782 -> 815,139 (-49.2%)
- collection-control versions/key: 1.982 -> 1.004
- total checkpoint time: 3,217.6 ms -> 2,949.9 ms (-8.3%)

The retained six-commit correctness gate also verified every commit and the final Git tree.

## Validation

- `cargo test -p lix_engine` (1,617 unit tests, 429 integration tests, 3 doctests)
- `cargo clippy -p lix_engine --all-features --all-targets -- -D warnings`
- focused checkpoint, partial-checkpoint, collection delete/recreate, protocol rejection, and 10k cascade-capacity regressions
- retained 6-commit and 50-commit all-plugin SlateDB replays with checkpoint-per-commit and Git-tree verification

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3355,10 +3355,12 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_json",
+ "slatedb",
  "smallvec",
  "tempfile",
  "tokio",
  "tracing-subscriber",
+ "ulid",
  "xxhash-rust",
 ]
 
--- a/packages/engine-benchmarks/Cargo.toml
+++ b/packages/engine-benchmarks/Cargo.toml
@@ -16,7 +16,7 @@ workspace = true
 default = ["rocksdb"]
 rocksdb = ["dep:lix_rocksdb_storage"]
 sqlite = ["dep:lix_sqlite_storage", "dep:rusqlite"]
-slatedb = ["dep:lix_slatedb_storage", "dep:object_store"]
+slatedb = ["dep:lix_slatedb_storage", "dep:object_store", "dep:slatedb", "dep:ulid"]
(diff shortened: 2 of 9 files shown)


samuelstroschein:
@codex review

PR #1010, opened 2026-07-31
Admit bounded large plugin entity pages

samuelstroschein:
## What changed

- scale plugin transition record/page admission for large file-derived entities while retaining the normal 2 MiB schedule for small files
- cap scaled pages at 16 MiB within the existing aggregate transition budget
- treat `next-page(max-bytes)` as an upper-bound hint and clamp to a source's smaller internal schedule
- apply the scaled bound to both cold and warm byte transitions

## Root cause

OpenClaw commit `c5c50a2141f2cdd805ae9b70a14a2e66dabac9b6` contains a 5,298,036-byte one-line source map. Git-text preserves that line byte-exactly as a base64 semantic snapshot (~7.06 MiB), which cannot fit the fixed 2 MiB push page. Cold successor hydration also rejected a consumer page hint larger than the source's internal schedule instead of returning a smaller valid page.

## Impact

Large, valid one-line text entities now cross the fused WASM push sink without weakening the aggregate byte budget. The same bounded policy covers initial open, cold successor, and later warm edits.

## Validation

- focused engine transition-limit and source-page tests
- focused rs-sdk binding admission test
- all 11 `plugin_git_text` unit tests
- release build with mold
- exact one-commit OpenClaw replay at `c5c50a2141f2cdd805ae9b70a14a2e66dabac9b6`, SlateDB, all WASM plugins, checkpoint every commit, per-commit state verification, and final Git-tree verification: 3,423 changed paths, success

### DIFF
--- a/packages/engine/src/plugin/incremental.rs
+++ b/packages/engine/src/plugin/incremental.rs
@@ -1554,12 +1554,10 @@ impl VecSourceState {
                 "component packet source page size must be positive",
             ));
         }
-        if requested > self.limits.max_page_bytes {
-            return Err(invalid_input(
-                "component packet source page request exceeds max_page_bytes",
-            ));
-        }
-        Ok(u64::from(requested))
+        // `max-bytes` is an upper bound supplied by the consumer, not an
+        // exact-size demand. A source with a smaller fixed schedule remains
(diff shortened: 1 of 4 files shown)


samuelstroschein:
@codex review

PR #1011, opened 2026-07-31
Bound binary replay memory and profile physical storage

samuelstroschein:
## What changed

- bound parent-tree Git/LFS reads and seed transactions by payload size instead of retaining an entire multi-gigabyte tree in memory
- bound final tree verification by both payload bytes and path count, while preserving full byte-length, SHA-256, mode, and Git OID verification
- report parent seed transaction counts in replay profiles
- add fast offline SlateDB SST-space profiling that works without decoding historical logical codecs
- declare the profiler example feature requirements explicitly

## Why

The real VS Code Docs Git LFS corpus materializes roughly 2.7 GiB of binary assets. The previous bootstrap and final verifier accumulated the whole tree at once and were killed by the OOM killer. Separately, logical layout inspection could not profile retained databases after codec changes. Both issues blocked data-driven physical-layout work.

## Evidence

The retained head-adjacent two-commit gate seeded 8,352 files and 2,724,026,017 unique Git LFS bytes in 22 bounded transactions, replayed/checkpointed 2/2 commits, and verified the final tree in 8.73 seconds. The retained root-history gate authenticated 1,898 unique LFS objects / 698,800,904 bytes across 2,308 changed paths.

Offline profiling of the retained failed OpenClaw checkpoint-every-commit run through about 511 commits attributed 4.62 MiB (17.4% of all data blocks and 44.8% of non-binary blocks) to commit delta segments/manifests, establishing the next optimization target.

## Validation

- `cargo test -p lix_cli git_replay --lib` (26 passed)
- `cargo clippy -p lix_cli --all-targets --no-deps -- -D warnings`
- `cargo clippy -p lix_engine_benchmarks --example storage_layout --features storage-benches,slatedb -- -D warnings`
- real VS Code Docs historical Git LFS two-commit gate
- real VS Code Docs head-adjacent 2.7 GiB bootstrap/checkpoint/final-verification gate

All Rust linking used mold.

### DIFF
--- a/packages/cli/src/commands/exp/git_replay.rs
+++ b/packages/cli/src/commands/exp/git_replay.rs
@@ -20,6 +20,10 @@ use zip::write::SimpleFileOptions;
 
 const PROGRESS_EVERY: usize = 10;
 const DEFAULT_INSERT_BATCH_ROWS: usize = 100;
+const TREE_BLOB_READ_BATCH_ROWS: usize = 8;
+const TREE_TRANSACTION_TARGET_BYTES: usize = 128 * 1024 * 1024;
+const FINAL_VERIFY_MAX_ROWS: usize = 512;
+const FINAL_VERIFY_TARGET_BYTES: usize = 128 * 1024 * 1024;
 // Four SHA-256 `<oid>\n` requests are 260 bytes, below POSIX `PIPE_BUF`.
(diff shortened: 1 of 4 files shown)


samuelstroschein:
@codex review

PR #1015, opened 2026-07-31
perf(engine): stream ordered commit deltas

samuelstroschein:
## What changed

- stream strictly ordered, fully addressable commit deltas into bounded 128-row segments
- retain only one candidate segment plus compact per-segment row counts instead of transaction-wide encoded arenas, sort tuples, and one UUID per row
- append sidecar payload encoding directly into the destination arena, removing a temporary allocation and copy per payload
- preserve the generic path for unordered, selected-change, and mixed-addressability commits

## Why

The retained 1M trace was OOM-killed after storage lowering because the prepared typed batch, commit materialization arenas, and adapter batch were simultaneously live. The existing generic commit-delta packer retained transaction-wide intermediates even though certified bulk inserts already arrive in strict storage order and every row has a direct address.

This specializes that certified shape at the shared transaction/storage boundary. It reduces peak lifetime overlap without introducing a SQL-only representation or changing the physical history format.

## Performance

Cached comparable create measurements for the combined stack versus merged main/#1009:

| Adapter / scale | main | this stack | gain |
| --- | ---: | ---: | ---: |
| RocksDB 10k | 36.820 ms | 29.083 ms | 21.0% |
| RocksDB 100k | 274.236 ms | 200.736 ms | 26.8% |
| SlateDB 100k | 263.893 ms | 198.652 ms | 24.7% |

The 100k traced commit materialization phase fell from 175 ms on merged main to 101 ms on this stack. Direct payload appends reduced the ordered-stream subphase from 44.1 ms to 36.4 ms.

This PR is stacked on #1012. The known incremental 10k RocksDB result over #1012 alone is 30.991 ms to 29.083 ms; the table above reports the review crossing against merged main so the >20% threshold is not obscured.

## Compatibility and correctness

- no storage-format change or backward shim
- a regression compares the streaming and generic manifest/segment key-value bytes exactly
- existing point replay, schema replay, GC, executeBatch, update/delete, and checkpoint coverage remains on the same physical representation

## Validation

- `cargo test -p lix_engine --all-features ordered_addressable_commit_delta_streams_segment_assignments -- --nocapture`
- `cargo test -p lix_engine --all-features tracked_state::storage::tests::packed_commit_delta -- --nocapture`
- `cargo test -p lix_engine --all-features execute_batch_lowers_distinct_bound_entity_inserts_once -- --nocapture`

### DIFF
--- a/packages/engine/src/tracked_state/mod.rs
+++ b/packages/engine/src/tracked_state/mod.rs
@@ -36,7 +36,7 @@ pub(crate) use storage::{
     scan_change_records_from_commit_deltas, scan_commit_delta_inventory, scan_commit_delta_values,
     stage_addressable_commit_deltas, stage_change_locators, stage_commit_deltas,
     stage_delete_change_locators, stage_delete_commit_delta_inventory_entry,
-    stage_delete_commit_roots,
+    stage_delete_commit_roots, stage_ordered_addressable_commit_deltas,
 };
 #[cfg(feature = "storage-benches")]
 pub(crate) use storage::{
--- a/packages/engine/src/tracked_state/storage.rs
+++ b/packages/engine/src/tracked_state/storage.rs
@@ -5,7 +5,7 @@
 )]
 
 use std::borrow::Cow;
-use std::collections::{BTreeMap, BTreeSet, HashMap};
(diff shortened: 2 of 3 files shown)


samuelstroschein:
@codex review

PR #1016, opened 2026-07-31
perf(runtime): compact certified packet key validation

samuelstroschein:
## Summary

- intern packet schema names once per certified page
- validate explicit keys as length-delimited BLAKE3 fingerprints instead of owned `Vec<String>` trees
- preserve generated-ID collision checks and ordinary/certified page duplicate semantics

## Why

The host validator decoded and retained every JSON entity key as owned strings before storage lowering. On the 10 MiB / 39,871-row JSON import, this made packet draining allocation-heavy even though storage consumes the certified payload directly.

## Benchmark

Current main → this PR, five-sample 10 MiB JSON push-sink import:

- p50: 291.701 ms → 269.725 ms (-7.5% end-to-end)
- packet drain: ~90 ms → ~67 ms (~-25%)
- host allocations: 412,644 → 80,409 (-80.5%)
- host allocated bytes: 126.56 MB → 116.63 MB (-7.8%)
- exact output: 39,871 semantic rows and source bytes preserved

Stacking the separate JSON fresh-open arena cut reaches 258.930 ms (-11.2% end-to-end); that plugin-specific change is intentionally excluded here.

## Validation

- `cargo test -p lix_sdk --lib` (29 passed)
- focused certified packet duplicate/collision regressions
- `git diff --check`
- workspace Clippy reaches only pre-existing dead-code denials in engine checkpoint and rs-sdk e2e helpers; no findings in this diff

### DIFF
--- a/packages/rs-sdk/src/default_wasm_runtime_component.rs
+++ b/packages/rs-sdk/src/default_wasm_runtime_component.rs
@@ -1864,8 +1864,20 @@ fn validate_typed_csv_quote_layout(quote_layout: &[u8], field_count: u16) -> Res
 }
 
 struct ValidatedCreatedPacketPage {
-    schemas: std::collections::BTreeSet<String>,
-    identities: Vec<CreatedPacketIdentity>,
+    schemas: Vec<String>,
+    identities: Vec<ValidatedPacketIdentity>,
+}
+
+enum ValidatedPacketIdentity {
+    Explicit {
+        schema_index: u32,
+        fingerprint: [u8; 32],
+        generated_local_ref: Option<u64>,
+    },
+    Create {
+        schema_index: u32,
+        local_ref: u64,
+    },
 }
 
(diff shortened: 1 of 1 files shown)


samuelstroschein:
@codex review

samuelstroschein:
Real-repo correctness gate on the hydrated vscode-docs fixture also passes exactly: 50/50 commits, 733 changed paths, checkpoint every commit, state verification every commit, and final Git tree verification.

This window is Markdown/text dominated and is not a useful positive performance signal for the JSON packet validator: replay-only was 48.042s main vs 50.017s candidate (+4.1%), while whole-process wall was 70.96s vs 69.77s and RSS 1,055,304 vs 1,053,176 KiB. The replay-only difference is concentrated in checkpoint/verify variance; execute was +2.1%. Treating this as a correctness/no-material-regression gate, while the isolated 10 MiB JSON lane remains the performance evidence.

PR #1017, opened 2026-07-31
perf(json): skip transient lookup on fresh imports

samuelstroschein:
## Summary

- distinguish fresh JSON imports from reparses used for identity reconciliation
- skip building the BLAKE3 identity lookup when the document has no predecessor
- retain the lookup for file-change reconciliation, hydration, and semantic edits

## Why

Fresh `open-file` parsed every node and then hashed every semantic identity into a lookup that cannot be consulted during that transition. The actor discarded that lookup after emitting the initial entities.

## Benchmark

Five-sample 10 MiB / 39,871-row JSON push-sink import, stacked on merged #1016:

- #1016 alone: 269.725 ms p50
- #1016 + this PR: 258.930 ms p50 (a further ~4.0%)
- original main before the two isolated cuts: 291.701 ms p50
- complete stack: -11.2% end-to-end
- guest linear-memory high water: 50,790,400 → 49,414,144 bytes (-2.7%)
- exact output bytes and 39,871 semantic rows preserved

A more invasive direct packet-streaming prototype was rejected: it improved the guest-open subphase but did not improve end-to-end time or memory.

## Validation

- `cargo test -p plugin_json` (6 passed)
- dedicated test proves fresh imports omit the lookup while reconcilable opens retain it
- exact 10 MiB benchmark validates output bytes and semantic rows
- `git diff --check`

### DIFF
--- a/plugins/json/src/core.rs
+++ b/plugins/json/src/core.rs
@@ -1110,9 +1110,30 @@ impl Document {
         _path: Option<&str>,
         namespace: IdNamespace,
     ) -> Result<(Self, InitialChanges), String> {
+        Self::open_parsed_file(bytes, namespace, true)
+    }
+
+    pub fn open_fresh_file(
+        bytes: Vec<u8>,
+        _path: Option<&str>,
+        namespace: IdNamespace,
+    ) -> Result<(Self, InitialChanges), String> {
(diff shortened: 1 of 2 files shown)


samuelstroschein:
@codex review

PR #1018, opened 2026-07-31
Compress inline binary history with bounded dictionaries

samuelstroschein:
## Summary

- encode small binary history as authenticated one-hop prefix/suffix deltas when a full inline base is available
- train one content-addressed 64 KiB zstd dictionary from a hard-bounded repository sample set
- keep sample state constant at 64 hash slots, consume the first 32 occupied slots once, and remove all sample rows after training
- cache native zstd dictionary contexts per write/read batch
- let WASM readers decode native dictionary-backed manifests with `ruzstd` 0.8.3; WASM never trains or writes dictionary samples
- hard-cut the repository protocol to `binary-history-dictionary.v29`
- extend storage-layout profiling with physical/logical binary-manifest and commit-delta source accounting

The delta base is always a full inline manifest, so reads remain one hop. Dictionary rows and reconstructed blobs are content-address authenticated.

## Data

OpenClaw replay with Markdown/CSV/text plugins and a checkpoint after every Git commit:

| Replay | Baseline | Candidate | Change |
|---|---:|---:|---:|
| 50 commits, DB bytes | 3,588,044 | 3,291,024 | -8.28% |
| 50 commits, replay | 1.152 s | 1.091 s | -5.3% |
| 500 commits, DB bytes | 26,174,143 | 23,458,765 | -10.37% |

The 500-commit candidate verified all 500 checkpoints and final state. Its raw replay was 25.555 s versus an older 24.152 s baseline, but 1.041 s of the 1.119 s checkpoint delta was one compaction/GC outlier at a different commit than the equivalent outlier in the prior candidate run. Excluding that unrelated event leaves about 1.5%; the baseline also predates the latest runtime merges, so this is not presented as a clean runtime comparison.

The measured dictionary setup hot path improved from 26.644 ms to 2.377 ms for 2,000 payloads (11.2x), and the 500-commit candidate spent about 350 ms less reading blobs.

Git LFS/binary coverage used two `vscode-docs` commits with 6,973 materialized LFS files totaling 2,724,026,017 bytes. All commits and the full final tree verified; the retained Lix DB is 2,587,692,411 bytes.

All replay databases were retained for storage inspection and follow-up queries.

## Validation

- `cargo check -p lix_engine --all-targets --features storage-benches`
- `cargo clippy -p lix_engine --all-targets --features storage-benches -- -D warnings`
- `cargo test -p lix_engine binary_cas:: --lib` (46 passed)
- `cargo test -p lix_engine merge --lib` (48 passed)
- `cargo test -p lix_engine diff --lib` (84 passed)
- `cargo test -p lix_engine checkpoint --lib` (17 passed)
- `RUSTFLAGS='--cfg getrandom_backend="wasm_js"' cargo check -p lix_js_sdk --target wasm32-unknown-unknown`

Native Rust validation used mold for linking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4670,9 +4670,9 @@ checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ruzstd"
-version = "0.8.2"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5ff0cc5e135c8870a775d3320910cd9b564ec036b4dc0b8741629020be63f01"
+checksum = "a7c1c839d570d835527c9a5e4db7cb2198683a988cb9d7293fc8674e6bd58fc8"
 
 [[package]]
 name = "ryu"
--- a/packages/engine-benchmarks/examples/storage_layout.rs
+++ b/packages/engine-benchmarks/examples/storage_layout.rs
@@ -4,7 +4,8 @@ use std::sync::Arc;
 
 use lix_engine::storage_adapter::{StorageAdapter, StorageReadOptions};
 use lix_engine::storage_bench::{
-    commit_delta_layout_accounting, layout_accounting, layout_space_catalog, space_inventory,
+    binary_manifest_layout_accounting, commit_delta_layout_accounting, layout_accounting,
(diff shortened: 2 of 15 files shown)


samuelstroschein:
@codex review

samuelstroschein:
@codex review

samuelstroschein:
@codex review

PR #1019, opened 2026-07-31
perf(csv): flatten cold import arena

samuelstroschein:
## Summary

- store cold-import CSV rows and fields in two flat arena arrays instead of allocating one field vector per row
- derive dense row IDs and canonical order ranks while emitting typed pages
- replace per-row 128-bit rank division with exact quotient/remainder accumulation
- skip quote-layout scans for rows already known to contain no quoted fields

## Why

The 220k-row cold path retained 220k independently allocated field vectors and performed a 128-bit division for every emitted row. Those structures are unnecessary because the initial importer is one-shot and rows are emitted in ordinal order.

## Benchmark

Seven-sample 10.68 MiB / 220,001-row typed CSV import on current main:

- elapsed p50: 136.136 ms → 119.799 ms (-12.0%)
- guest linear-memory high water: 36,962,304 → 28,573,696 bytes (-22.7%)
- one top-level guest export
- 68 bounded 256 KiB typed pages
- zero actor worker threads
- exact row count plus reopen/query validation preserved

The first sample includes component compilation and is reported separately by the benchmark; p50 uses the complete seven-sample set.

## Validation

- `cargo test -p plugin_csv` (7 passed)
- release regression emits all 220,001 rows across page boundaries without rank drift/overflow
- flat scanner parity test covers quoted fields, escaped quotes, delimiters, LF, CRLF, and CR
- `cargo clippy -p plugin_csv --all-targets -- -D warnings`
- exact 10.68 MiB benchmark and reopened semantic query
- `git diff --check`

### DIFF
--- a/plugins/csv/src/core.rs
+++ b/plugins/csv/src/core.rs
@@ -2800,9 +2800,25 @@ impl Document {
 #[allow(dead_code)]
 pub struct ColdInitialImport {
     bytes: Vec<u8>,
-    rows: Vec<RowDraft>,
+    rows: Vec<ColdRowDraft>,
+    fields: Vec<FieldRange>,
     dialect: Dialect,
     next_row: usize,
+    next_order_rank: u64,
+    order_remainder: u64,
+    order_step: u64,
+    order_remainder_step: u64,
+    order_denominator: u64,
+}
+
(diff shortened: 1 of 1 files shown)


samuelstroschein:
@codex review

PR #1021, opened 2026-07-31
perf(engine): stream typed creates and batch history reads

samuelstroschein:
## Summary

- hard-cut tracked commit deltas to `LXCD8` / change-locator v2, with 512-row ordered create segments and 128-row generic history segments
- lower eligible `executeBatch` INSERT values directly into the shared typed transaction batch instead of constructing an Arrow batch, then transfer certified raw columns into prepared state without cloning identities or re-interning dictionaries
- reuse large ordered commit deltas as packed current-state bases
- resolve direct change IDs in physical batches and decode/decompress each selected history segment once, preserving explicit address-shaped ID fallback semantics
- update the tracked working-diff benchmark to the current `diff_type` API

This stays below SQL-specific JSON materialization: the typed transaction representation is the shared producer/storage boundary. Plugin execution remains out of scope. The format/locator changes are intentional hard cuts with no compatibility shim.

## Performance

Exact PR #1015 baseline vs this branch, 100k create:

| Adapter | Baseline | This branch | Change |
| --- | ---: | ---: | ---: |
| RocksDB | 200.736 ms | 158.091 ms | **-21.25%** |
| SlateDB | 198.652 ms | 153.283 ms | **-22.84%** |

The retained 1M create now completes instead of being OOM-killed:

- RocksDB: 3.642 s
- SlateDB: 3.627 s

History non-regression, exact PR #1015 baseline vs this branch:

| Operation | RocksDB | SlateDB |
| --- | ---: | ---: |
| Merge preview | 209.550 → 29.256 ms (**-86.0%**) | 202.879 → 32.750 ms (**-83.9%**) |
| Merge commit | 42.500 → 15.181 ms (**-64.3%**) | 45.750 → 19.369 ms (**-57.7%**) |
| Working diff | 0.938 → 0.942 ms (flat) | 1.708 → 0.717 ms (**-58.0%**) |
| Checkpoint | 12.817 → 10.782 ms (**-15.9%**) | 14.891 → 13.941 ms (**-6.4%**) |

CRUD read/update/delete checks are flat-to-faster; `delete_all` improved about 20%.

## Validation

- `cargo test -p lix_engine --lib --all-features` — 1641 passed
- `cargo clippy -p lix_engine --all-targets --all-features -- -D warnings`
- `cargo check -p lix_engine --all-features`
- `cargo test -p lix_engine_benchmarks --test tracked_state_crud_public_result --features storage-benches,slatedb` — 4 passed
- `cargo fmt --all -- --check`
[...]

PR #1024, opened 2026-07-31
perf(engine): stream certified parameter inserts

samuelstroschein:
## Summary

- construct fixed-shape certified parameter INSERT batches directly from typed identity and snapshot columns, including a single-column identity path with no component scratch buffer
- stage this exact active-branch tracked shape without generic plugin reconciliation, storage-scope scans, filesystem preflight, or per-row commit-id writes
- carry a strict physical-key-order certificate into commit lowering so staging and addressable history publication can skip redundant O(rows) proof scans
- represent first-batch INSERT statement attribution as implicit row ordinals, materializing an explicit column only if later transaction writes require it
- preserve fallback attribution for unordered/irregular batches and consecutive explicit-transaction batches

This is stacked on #1021. Plugin execution remains out of scope. The experimented 1,024/2,048-row history granules are deliberately **not** included; this PR retains #1021's 512-row / 64 KiB history layout after exact-read controls showed that write granularity must be evaluated separately.

## Performance

Exact #1021 baseline versus this branch, tracked SQL `executeBatch` create:

| Rows | Adapter | #1021 | This branch | Change |
| ---: | --- | ---: | ---: | ---: |
| 100k | RocksDB | 158.091 ms | 130.483 ms | **-17.5%** |
| 100k | SlateDB | 153.283 ms | 125.916 ms | **-17.9%** |
| 1M | RocksDB | 3.642 s | 2.746 s | **-24.6%** |
| 1M | SlateDB | 3.627 s | 2.798 s | **-22.9%** |

The 1M runs complete without OOM at about 2.31 GiB RocksDB / 2.35 GiB SlateDB peak RSS.

The exact local #1021 control and candidate retain the same physical history layout. Repeated working-diff runs were flat within run variance (RocksDB 1.221 vs 1.258 ms; SlateDB 1.154 vs 1.162 ms). Checkpoint was faster on RocksDB (12.450 to 6.036 ms) and flat on SlateDB (12.802 vs 13.163 ms). Merge/diff/checkpoint correctness remains covered by the full engine suite.

At 10k rows the current end-to-end create path remains about 2.5x raw SQLite (20.50 ms RocksDB / 20.04 ms SlateDB versus 8.24 ms SQLite in the retained measurement), so this PR is one stacked layer rather than completion of the 1.2x objective.

## Correctness

- certified ordering is produced from the already validated typed primary-key stream and is cleared by every topology-changing batch operation
- uncommon fallback paths reconstruct explicit statement indices
- consecutive certified batches preserve the failing batch's local `statementIndex`
- a later generic insert cannot inherit an implicit ordinal from an earlier certified batch
- collection-generation replacement semantics remain checked before direct staging

## Validation

- `cargo test -p lix_engine --lib --all-features` — 1,643 passed, 6 ignored
- `cargo clippy -p lix_engine --all-targets --all-features -- -D warnings`
- `cargo test -p lix_engine_benchmarks --test tracked_state_crud_public_result --features storage-benches,slatedb` — 4 passed
- focused certified batch, consecutive-batch attribution, generic-after-certified attribution, and ordered addressable history tests

### DIFF
--- a/.changenotes/stream-certified-parameter-inserts.md
+++ b/.changenotes/stream-certified-parameter-inserts.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+Large bound tracked-entity INSERT batches now stage certified typed columns directly and avoid repeated per-row transaction bookkeeping.
+
+Million-row creates use less peak memory and complete more than 20% faster on both RocksDB and SlateDB while retaining the existing history page granularity.
--- a/packages/engine/src/entity_pk.rs
+++ b/packages/engine/src/entity_pk.rs
@@ -201,6 +201,29 @@ impl EntityPk {
         )
     }
 
+    /// Builds an all-string identity after a typed producer has validated the
+    /// component count and scalar types against the active schema plan.
+    pub(crate) fn from_validated_shared_string_parts(
+        parts: impl IntoIterator<Item = SharedStr>,
+    ) -> Self {
+        let components = parts
+            .into_iter()
+            .map(EntityPkComponent::String)
+            .collect::<EntityPkComponentBuffer>();
+        debug_assert!(!components.is_empty());
+        Self {
+            components: EntityPkComponents::from_smallvec(components),
+        }
+    }
+
+    /// Builds the dominant one-column string identity without assembling an
+    /// intermediate component buffer after typed ingress validation.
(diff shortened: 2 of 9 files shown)


samuelstroschein:
@codex review

PR #1023, opened 2026-07-31
alias complete checkpoint commit deltas

samuelstroschein:
## What changed

- represent complete single-source checkpoint selections as a certified source-commit alias plus the checkpoint's local authored overlay
- persist an order-independent member count/fingerprint certificate in every commit-delta manifest so checkpoint admission is O(selected rows) without scanning the source segment
- teach point history, schema scans, root rebuild, inventory, canonical-history, and GC paths to resolve the compact checkpoint representation
- hard-cut the repository protocol to `checkpoint-source-delta.v33`; no migration or backward compatibility is provided
- restrict alias admission to checkpoint-marker commits and certify identity, change ID, deletion state, and timestamps

## Why

Checkpointing every replayed Git commit previously re-encoded and persisted the selected semantic rows even when the selection was exactly one immutable source commit. OpenClaw profiling showed 496/497 eligible checkpoints in the first 500 commits. The compact representation removes that checkpoint CPU/write amplification while preserving eager semantic materialization and WASM plugins.

## Measured impact

Release-mode OpenClaw replay, all plugins enabled, SlateDB, checkpoint every commit, state verification enabled:

| 500 commits | exact parent | candidate | change |
| --- | ---: | ---: | ---: |
| replay elapsed | 29.89 s | 26.38 s | -11.7% |
| checkpoint phase | 4.14 s | 2.55 s | -38.5% |
| verify phase | 5.55 s | 5.43 s | -2.2% |

The retained candidate compacted 418 checkpoint manifests covering 43,513 selected rows into about 105 KiB of local manifests. Compressed commit-history SST bytes changed only about -0.5% because SlateDB already compresses adjacent duplicate checkpoint rows efficiently; this PR is a checkpoint CPU/write-amplification optimization, not a disk-capacity claim.

Retained evidence:

- `/root/projects/openclaw-perf-data/openclaw-v35-checkpoint-certificate-500-candidate-db`
- `/root/projects/openclaw-perf-data/openclaw-v35-checkpoint-certificate-500-candidate-profile.json`
- `/root/projects/openclaw-perf-data/openclaw-v33-exact-parent-500-baseline-db`
- `/root/projects/openclaw-perf-data/openclaw-v33-exact-parent-500-baseline-profile.json`

## Validation

- OpenClaw: 500/500 commits verified plus final Git tree
- post-fix release protocol gate: 6/6 commits verified plus final Git tree
[...]

PR #1020, opened 2026-07-31
perf(plugin): retain host-validated entity authority

samuelstroschein:
## What changed

- retain host-validated entity identity authority with each live plugin actor
- represent dense certified create identities as compact UUID namespace ranges
- use that authority to bypass redundant exact durable reads for keyed successor changes
- keep sparse insert/delete overlays persistent and publish them atomically with the successor document

## Why

A tiny edit to the 1.24 MiB `vscode-api.md` document was materializing the complete 3,808-row certified segment merely to prove that one keyed Markdown block already existed. The actor already owned a validated semantic document, but discarded the host's proof of which creatable identities belonged to it.

This retains only host-proven authority; unknown identities still take the durable exact-read path. Guest-provided identifiers are never trusted as authority.

## Performance

Large vscode-api Markdown exact transition, current main baseline → candidate:

- p50: 8.979 ms → ~6.9 ms (about 23% faster)
- cumulative host allocation: 40.77 MB → 22.27 MB (about 45% less)
- `plugin_create_rows`: 1.2–2.0 ms → ~0.006 ms
- exact semantic output and all 3,808 rows preserved

Cross-format guard lanes:

- 10.68 MiB / 220,001-row typed CSV: 119.80 ms p50
- 10 MiB JSON sparse successor: 8.42 ms p50
- 1.85 MiB / 20,000-element Excalidraw successor: 3.98 ms p50

## Validation

- `cargo check -p lix_sdk`
- focused actor authority publication, compaction, and range override tests
- focused certified create-range runtime test
- `cargo clippy -p lix_engine -p lix_sdk --all-targets`
- 3-commit / 26-path real Lix repository replay with all WASM plugins, checkpoint every commit, state verification every commit, and final Git tree verification

PR #1025, opened 2026-07-31
Store immutable binary chunks once

samuelstroschein:
## Summary

- store immutable content-addressed binary chunk payloads once in the backing object store
- retain zero-byte SlateDB marker rows for transactional publication, snapshot visibility, range indexing, and key-only reads
- hydrate exact full-value reads, scans, and value preconditions from the immutable object
- use a bounded zstd-3 envelope only when it saves at least 128 bytes
- account for immutable objects in the physical storage profiler
- hard-cut the repository protocol marker because the physical representation changed

This was developed on #1023, which is now merged into `main`.

## Why

Profiling a retained `vscode-docs` Git LFS replay showed binary CAS chunks dominated physical storage and more than 25% of sampled seed CPU was in zstd compression. Immutable content-hash-keyed payloads were being written through the WAL, memtable, SSTs, and compaction even though they are never updated in place.

The new layout publishes the payload object before its transactional marker. Rollback creates no object, marker visibility remains atomic, key-only reads remain payload-free, and old snapshots remain valid because immutable objects are never removed beneath visible markers. Growth is linear in unique binary content instead of bytes rewritten by LSM maintenance.

## Retained replay evidence

All measurements use the same pre-plugin-#1020 code, all Wasm format plugins, and checkpoint every Git commit.

### vscode-docs with Git LFS (2 commits, 6,973 LFS files / 2.724 GB materialized)

| phase | control | candidate | change |
| --- | ---: | ---: | ---: |
| seed | 95.120 s | 50.704 s | -46.7% |
| replay | 3.411 s | 2.006 s | -41.2% |
| final verification | 12.149 s | 7.353 s | -39.5% |
| measured total | 110.784 s | 60.202 s | -45.7% |

Physical bytes were effectively flat: 2,585,844,965 -> 2,585,146,948 (-0.03%). The candidate consists of 211,500,876 SST bytes plus 2,373,646,072 immutable-object bytes. All 2 commits and the final Git tree verified.

### OpenClaw (500 commits)

- replay: 26.378 s -> 26.331 s (-0.18%)
- final verification: 128.4 ms -> 101.9 ms
- 500/500 commits and final Git tree verified
- tiny-dataset apparent bytes: 21.16 MB -> 21.56 MB (+1.9%), attributable to filesystem overhead for 26 sidecar objects; this overhead is diluted on the asset-scale corpus

Retained databases and JSON profiles remain available on the benchmark host for storage queries and follow-up profiling.

## Correctness and failure semantics

- immutable writes are staged in the transaction and uploaded only during commit
- a marker is never published before its payload object exists
- failed multi-object uploads remove objects created by that failed batch
- create-if-absent makes retries and cross-commit dedupe idempotent
- deletes remove snapshot-visible markers but retain payloads required by older snapshots
- full-value equality/hash preconditions inspect the real immutable bytes
- missing published objects fail as storage I/O corruption rather than returning marker bytes

## Validation

- `git diff --check`
- retained 500-commit OpenClaw replay with per-commit checkpoints and verification
- retained 2-commit `vscode-docs` Git LFS replay with 6,973 materialized LFS objects and verification

Rust linking for local validation used mold.

### DIFF
--- a/.changenotes/store-immutable-binary-chunks-once.md
+++ b/.changenotes/store-immutable-binary-chunks-once.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+Large immutable binary chunks now bypass repeated LSM-tree rewriting while retaining transactional visibility and snapshot-safe reads.
+
+Git LFS replay stores content-addressed chunk payloads once in the backing object store and keeps only compact transactional markers in SlateDB.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3529,6 +3529,7 @@ dependencies = [
  "slatedb-common",
  "tempfile",
  "tokio",
+ "zstd",
 ]
 
 [[package]]
--- a/packages/engine-benchmarks/examples/storage_layout.rs
+++ b/packages/engine-benchmarks/examples/storage_layout.rs
@@ -12,7 +12,7 @@ use object_store::local::LocalFileSystem;
 use slatedb::{SstReader, ValueDeletable};
 
 const HOT_SPACE: &str = "live_state.hot_row.v20";
-const IMMUTABLE_BINARY_CAS_CHUNK_DIR: &str = "db/lix-immutable-binary-cas-chunk-v2";
+const IMMUTABLE_BINARY_CAS_CHUNK_DIR: &str = "db/lix-immutable-binary-cas-chunk-v3";
 
 #[derive(Default)]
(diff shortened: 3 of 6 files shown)


samuelstroschein:
@codex review

samuelstroschein:
@codex review

samuelstroschein:
@codex review

PR #1026, opened 2026-07-31
fix(engine): cut repository protocol for LXCD8

samuelstroschein:
Advances the repository protocol marker to V34 for the LXCD8 packed commit-delta format and adds a regression proving V33 repositories fail closed before LXCD7 locators are decoded.

Root cause: #1021 widened segment coordinates and changed the format magic without changing the repository-wide compatibility marker.

Validation:
- cargo fmt --all
- git diff --check
- production lix_engine compilation completed locally
- focused test linking and test-metadata compilation exceeded the constrained local RAM ceiling and were SIGKILLed; CI is the execution gate

### DIFF
--- a/packages/engine/src/engine.rs
+++ b/packages/engine/src/engine.rs
@@ -791,6 +791,30 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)


samuelstroschein:
@codex review

PR #1027, opened 2026-07-31
Share SlateDB disk cache budget

samuelstroschein:
## Summary

- treat `max_disk_cache_bytes` as one total allowance for SlateDB SST objects and immutable binary payloads
- split that allowance exactly between the two disk caches without exceeding odd-byte totals
- prune existing immutable cache contents to their new share when reopening

## Root cause

#1025 added a separate immutable-payload disk cache but gave it the full configured allowance while SlateDB's object cache independently retained the same full allowance. A workload filling both caches could therefore consume roughly twice the requested disk capacity.

## Validation

- focused shared-budget tests cover odd and even totals
- focused remote immutable-cache test still serves payload bytes after authoritative remote removal
- immutable cache bound test covers pruning
- `cargo clippy -p lix_slatedb_storage --all-targets --no-deps -- -D warnings`
- `git diff --check`

Local Rust validation used mold.

### DIFF
--- a/.changenotes/share-slatedb-disk-cache-budget.md
+++ b/.changenotes/share-slatedb-disk-cache-budget.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+SlateDB now keeps its table cache and immutable binary payload cache within one configured disk-cache limit.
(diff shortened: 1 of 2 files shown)


samuelstroschein:
@codex review

PR #1028, opened 2026-07-31
Reuse existing CAS manifests for prehashed blobs

samuelstroschein:
## What changed

When a binary CAS payload already carries a trusted content hash, check for an existing manifest before preparing and probing every chunk. If the manifest exists and its size matches, return its receipt without staging any writes.

## Why

Immutable auxiliary payloads can recur across transactions. Their content hash already identifies the exact durable blob, so rechunking and probing all chunks adds CPU, allocation, and storage reads without changing the write set.

## Validation

- `cargo test -p lix_engine prehashed_existing_blob_reuses_manifest_without_staging_chunks`
- The focused regression commits a multi-chunk blob, restages the same prehashed payload, and verifies an identical receipt plus an empty second write set.

### DIFF
--- a/packages/engine/src/binary_cas/kv.rs
+++ b/packages/engine/src/binary_cas/kv.rs
@@ -3372,8 +3372,8 @@ mod tests {
 
         assert_eq!(
             writes.stats().staged_puts,
-            1,
-            "an inline repeat write should deterministically replace one manifest"
+            0,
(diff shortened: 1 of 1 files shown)


samuelstroschein:
@codex review

samuelstroschein:
@codex review

PR #1031, opened 2026-07-31
perf(engine): certify dense commits and compile typed insert layouts once

samuelstroschein:
## What changed

- hard-cut tracked commit-delta storage to LXCD9 / repository protocol V37
- replace ordered commits' per-row selection fingerprint with a dense `(segment, ordinal)` inventory
- certify checkpoint source membership at the immutable diff-reader boundary and prove exact selections from direct change coordinates
- compile typed SQL insert schema layouts once per parameter batch instead of repeating fixed property/required/primary-key lookups per row
- serialize canonical strings directly when the UTF-8 requires no JSON escaping, with the serde path retained for escaped values
- keep generic/unordered commit selections on the existing multiset fingerprint path

There is deliberately no backward shim. LXCD8 repositories fail closed and must be recreated.

## Why

The ordered create path was hashing every selected row after it had already assigned a direct immutable address. At the SQL boundary, the certified parameter batch then repeated structural schema work for every row. Together these costs were material at 10k and scaled linearly at 100k.

## Performance

Paired SQL-session `insert_all` profile, 15 independent samples, same host and exact current-main control binary:

| Rows | Adapter | Main median | PR median | Change |
| ---: | --- | ---: | ---: | ---: |
| 10k | RocksDB | 26.469 ms | 21.534 ms | -18.6% |
| 10k | SlateDB | — | 19.291 ms | — |
| 100k | RocksDB | 156.245 ms | 122.839 ms | -21.4% |
| 100k | SlateDB | — | 118.175 ms | — |

The 100k trace identified batch certification as the next scaling cost; compiling the layout once is what takes the combined stack over the 20% publication threshold at scale.

Focused history controls remained noise-scale or improved. Repeated checkpoint medians were 19.430 → 20.275 ms on RocksDB and 15.083 → 14.437 ms on SlateDB. Merge preview was 137.919 → 139.086 ms on RocksDB and 129.706 → 125.283 ms on SlateDB; committed merge was 14.227 → 15.460 ms and 19.628 → 18.550 ms respectively. Working-diff scans remained within single-digit variance.

## Validation

- `cargo test -p lix_engine --all-features`
  - 1,664 unit tests passed
  - 790 integration tests passed, 2 ignored
  - 3 doctests passed
- focused dense selection, canonical history alias, checkpoint, typed schema, canonical string escaping, and certified parameter-batch regressions
- feature-complete RocksDB + SlateDB release benchmarks linked with mold and at most two compiler jobs

### DIFF
--- a/.changenotes/dense-commit-selection-certificates.md
+++ b/.changenotes/dense-commit-selection-certificates.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-Tracked history now certifies ordered bulk commits from their dense address inventory instead of hashing every row independently.
+Tracked history now certifies ordered bulk commits from their dense address inventory instead of hashing every row independently. Typed SQL parameter batches also compile their fixed schema layout once and serialize escape-free strings directly.
--- a/.changenotes/dense-commit-selection-certificates.md
+++ b/.changenotes/dense-commit-selection-certificates.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+Tracked history now certifies ordered bulk commits from their dense address inventory instead of hashing every row independently.
--- a/packages/engine/src/catalog/mod.rs
+++ b/packages/engine/src/catalog/mod.rs
@@ -11,5 +11,5 @@ pub(crate) use schema::{
 };
 pub(crate) use snapshot::{
     CatalogFingerprint, CatalogSnapshot, DefaultPlan, StateDeleteReferencePlan, TransactionCatalog,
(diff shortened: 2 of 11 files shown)


samuelstroschein:
@codex review

samuelstroschein:
Fixed the timestamp-alias P1 in da45d58c4. Checkpoint batches now retain source-membership certification only when every staged tuple is source-exact; Added rows whose created_at is canonicalized to updated_at fall back to materialized checkpoint deltas. Added a regression and reran all focused checkpoint tests.

@codex review

PR #1074, opened 2026-08-01
perf(engine): index certified packet identity filters

samuelstroschein:
## Summary

- index requested semantic entity identities once per certified packet batch
- replace the per-row linear filter scan with constant-time membership checks
- document the historical-query performance improvement in the changelog

## Profile and cause

A 10-second `perf` sample while activating a historical Wesnoth branch attributed 87.16% of samples to `EntityPkComponents::eq` and 10.82% to `decode_certified_packet_rows`. The request contained tens of thousands of entity identities, and the decoder linearly rescanned that complete filter for every packet row.

This changes the work from O(packet rows × requested identities) equality checks to O(requested identities) index construction plus O(packet rows) lookups.

## Validation

- `cargo test -p lix_engine --features all-simulations` (1,674 tests across simulations plus doctests)
- `cargo clippy -p lix_engine --all-targets --features all-simulations -- -D warnings` reaches an unrelated pre-existing `dead_code` failure for `StorageWriteSet::apply`; no warning is introduced by this patch

The end-to-end benchmark rerun will be attached after exercising this commit together with the two other isolated benchmark fixes.

### DIFF
--- a/.changenotes/index-certified-packet-identity-filters.md
+++ b/.changenotes/index-certified-packet-identity-filters.md
@@ -1,5 +1,5 @@
 ---
-"@lix-js/sdk": patch
+type: patch
 ---
 
 Speed up historical semantic-state queries by indexing requested entity identities instead of rescanning the complete filter for every certified packet row.
--- a/.changenotes/index-certified-packet-identity-filters.md
+++ b/.changenotes/index-certified-packet-identity-filters.md
@@ -1,0 +1,5 @@
+---
+"@lix-js/sdk": patch
+---
+
+Speed up historical semantic-state queries by indexing requested entity identities instead of rescanning the complete filter for every certified packet row.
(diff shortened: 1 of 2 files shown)


samuelstroschein:
Benchmark follow-up after the initial profile:

- The first 10 s sample on `main` attributed 87.16% of cycles to `EntityPkComponents::eq`.
- Building an index per certified batch simply moved the hotspot to repeated `HashMap::insert`/cloning, so the final patch builds entity, file, and schema indexes once per scan request and shares them across batches.
- The full engine simulation suite passes and all five PR checks are green.
- In the final 3-repetition benchmark, commit listing is 0.08 s warm / 0.32 s cold on Wesnoth. Exhaustive all-plugin historical branch reconstruction still exits 137 at a median 18,924,488 KiB RSS; no-plugin succeeds in 6.50 s. That remaining eager snapshot-materialization/OOM is separate from the linear filter-scan CPU bug fixed here and is called out explicitly in the benchmark report.

Raw `perf.data` and text reports for the original linear scan and the intermediate per-batch index are included with the benchmark artifacts.

PR #1075, opened 2026-08-01
perf(engine): bound certified history hydration by commit

samuelstroschein:
## Summary

- group pending certified payload references by their owning commit
- scan and consume one commit's certified rows at a time
- avoid materializing union-filter matches from every requested commit simultaneously

## Why

Historical branch activation can request certified semantic payloads spanning many commits. The previous code formed one union of every schema, entity, and file key and scanned all commit batches into one `rows` vector before selecting exact `(commit, schema, file, entity)` payloads. On the 28,873-file Wesnoth benchmark this all-plugin path reproducibly reached about 18.9 GB RSS and exited 137, while no-plugin history succeeded.

Grouping by commit preserves the exact lookup and error behavior while bounding both the scan request and temporary payload map to one commit.

## Validation

- `cargo test -p lix_engine --features all-simulations` (1,674 tests across simulations plus doctests)

An end-to-end historical Wesnoth profile against the isolated integration build will be added after CI/build validation.

### DIFF
--- a/.changenotes/group-certified-hydration-by-commit.md
+++ b/.changenotes/group-certified-hydration-by-commit.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+Bound historical semantic payload hydration by querying certified rows one commit at a time instead of materializing cross-commit filter matches together.
--- a/.changenotes/group-certified-hydration-by-commit.md
+++ b/.changenotes/group-certified-hydration-by-commit.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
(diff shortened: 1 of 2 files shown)


samuelstroschein:
Closing after end-to-end profiling showed this grouping approach does not resolve the historical semantic hydration failure. On the 25-commit Wesnoth benchmark it was still killed by OOM after 2m42s at 17,954,968 KiB peak RSS. The profile moved time into payload materialization/allocation and large memory movement, so this is not a review-worthy fix on its own. I am keeping the certified-filter indexing fix isolated in #1074 and documenting historical hydration streaming/deduplication as the remaining product gap.

PR #1147, opened 2026-08-02
Preserve foreground latency under background load

### LINKED ISSUE #1133: Preserve foreground query latency under concurrent background workloads
## Problem

A background query loop against one Lix materially increases the tail latency of unrelated foreground queries against the same Lix.

This is independent of any plugin or Wasm runtime. In a native control benchmark, one background worker repeatedly executed database work while another client measured 1,000 normal point queries:

| metric | baseline | under native background load |
| --- | ---: | ---: |
| median p95 across 3 isolated runs | 333.5 µs | 438.7 µs |
| median p95 regression | — | **24.6%** |
| worst observed regression | — | **47.0%** |

The benchmark ran on Linux x86-64, AMD EPYC-Genoa, against the Rust SDK 0.9 engine. Each repetition ran in a fresh worker process.

Background sync, indexing, imports, compaction, agents, and maintenance jobs can all create this workload. A caller currently has no general way to mark work as background, bound its concurrency, or prevent it from degrading interactive requests.

## Expected direction

Lix should provide workload isolation or fair scheduling across independent sessions/clients. Possible mechanisms include:

- bounded admission queues instead of unbounded contention;
- per-session or per-principal concurrency limits;
- foreground/background priorities or weighted fair scheduling;
- cancellation propagated into active DataFusion work;
- queue-depth, wait-time, execution-time, and cancellation telemetry.

The API should be runtime-neutral and usable by any background integration.

## Acceptance criteria

- Add a deterministic native benchmark with one saturating background worker and one foreground point-query client.
- Report baseline and loaded p50/p95/p99 plus queue wait separately from execution time.
- Keep foreground p95 regression below 20% under the defined load, or document and enforce an explicit service-level policy.
- Demonstrate that cancelling or closing the background client releases queued and active work promptly.
- Avoid starvation in both directions: foreground work remains responsive and background work continues making progress.

## Related

#1039 tracks fixed-cost point-operation overhead. This issue is specifically about **concurrent workload isolation and tail latency**, not single-query speed.

samuelstroschein:
## Summary

Closes #1133.

Adds engine-scoped workload admission so callers can mark SQL statements and atomic batches as `background`. Foreground execution keeps an atomic-only fast path while background work is FIFO, bounded to 64 queued calls, and limited to one active statement/batch per engine.

The priority is wired through Rust, N-API, WASM, browser, and remote clients. Native N-API background work runs through a dedicated child-session actor so the main actor remains available to foreground calls. Queue wait, execution time, depth, and priority are emitted separately in SQL telemetry; saturation maps to retryable HTTP 429.

## Performance evidence

A deterministic native harness runs 21 fresh alternating baseline/default/priority triples, pins foreground and background workers to separate CPUs, records 2,000 foreground point queries per process, and uses a seeded 20,000-resample paired bootstrap.

| mode | foreground p50 | foreground p95 | foreground p99 | background queries/s |
| --- | ---: | ---: | ---: | ---: |
| baseline | 498.8 us | 550.1 us | 621.2 us | 0 |
| loaded-default | 647.5 us | 735.9 us | 899.4 us | 2,276.3 |
| loaded-priority | 512.2 us | 618.9 us | 678.5 us | 161.4 |

- Median paired p95 regression: 29.7% unclassified, 11.2% priority
- Priority bootstrap 95% interval: 5.7–13.8%
- Background queue p50/p95: 5.150/5.194 ms
- Background execution p50/p95: 1.042/1.318 ms
- Cooperative shutdown median: 3.087 ms
- Background progress occurred in every loaded process

This is an explicit QoS tradeoff: saturated background throughput falls by about 93% to protect foreground tail latency.

## Correctness and lifecycle

- bounded FIFO admission with per-waiter notification
- foreground is never queued behind forced background progress
- suppression deadline prevents background starvation
- dropping queued or active permits releases capacity
- session close cancels queued background acquisition
- native close rejects/drains queued requests and waits for active work
- branch switching closes the background child first, avoiding stale branch state
- explicit transactions reject background priority instead of silently ignoring it

Scheduling is cooperative at statement and batch boundaries; an already-running long SQL statement is not preempted. Coordination is per engine, not process-wide.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- Cargo checks for engine/SDK/server protocol
- workload admission, cancellation, transaction, telemetry, and HTTP mapping tests
- full JS SDK native + WASM + TypeScript + bundled WASI plugin build
- TypeScript typecheck
- 107 selected JS integration/remote tests
- release benchmark build, 21-triple suite, and sampled profile

## Review

Independent sub-agent reviews covered:
- concurrency/lifecycle architecture
- benchmark statistics and profiling methodology
- prior-art/API semantics and branch isolation

All three approved the final revision with no blocker or high-severity findings.


samuelstroschein:
Coordination from the generalized TPC-H/OLAP goal:

- merged #1145: generic statement-scoped exact scan reuse; matched TPC-H Q15 improved RocksDB 35.6% and SlateDB 35.0%. This touches SQL scan execution and should be included in your CRUD SELECT benchmarks.
- merged #1148: two boxed transaction-batch poll boundaries; minimum passing stack 2288 -> 1936 KiB (-15.38%), default 2 MiB now passes. No ordering/atomicity/read-your-writes changes.
- open #1146: generic arithmetic UPDATE (`+ - * / %`) with atomic simultaneous pre-image evaluation and exact staged post-image RETURNING. Literal UPDATE fast path remains unchanged; this may overlap only if your columnar literal-input branch touches shared entity provider routing.
- current profile: Q8 packed source reads ~50.2 MB compressed into ~177.2 MB Arrow; checksum+decompress+Arrow reconstruction is ~68.1 ms/run versus ~10.9 ms RocksDB materialization. We are targeting generic packed numeric decode/layout next, not SELECT shape recognition.

Please benchmark merged #1145/#1148 plus #1146 once green against your RocksDB+SlateDB 10k/1M CRUD suite and flag provider/layout conflicts here.

samuelstroschein:
Closing this prototype intentionally. The benchmark demonstrated that admission control can reduce foreground p95 regression under synthetic saturated background load (29.7% to 11.2%), but it also reduced background throughput by roughly 93% and introduced a cross-language public per-query priority API without a demonstrated production requirement.

If workload isolation becomes necessary, the better boundary is likely host-assigned, session-scoped classification for services/plugins, combined with resource-aware admission. The benchmark and implementation remain available in this branch/PR as prior art.

Issue #1144, opened 2026-08-02
Consume certified entity columnar carriers without duplicating row coordinates


samuelstroschein:
## Problem

During complete certified replacement commit, `prepare_entity_columnar_write_sets` borrows `CertifiedEntityColumnarBatch`, clones the encoded set, and copies every `RowGroupRowLocation` into `EntityColumnarWriteSets::state_row_locations`.

The encoded payload clone is cheap because columns are `Bytes`-backed, but at 1M rows the coordinate column exists twice during the peak commit window. `EntityColumnarWriteSets` also converts dense coordinates into `Vec<Option<RowGroupRowLocation>>`, adding unnecessary representation overhead for a proven dense replacement.

## Proposed direction

- Add a consuming/take path from `PreparedStateBatch` for the certified columnar carrier.
- Let `EntityColumnarWriteSets` adopt a dense coordinate vector directly (for example, a dense/sparse location representation) instead of copying into `Vec<Option<_>>`.
- Move the encoded set rather than cloning it.
- Verify identical storage writes and semantics; benchmark 1M UPDATE RSS and allocations on RocksDB and SlateDB.

This is a non-blocking memory follow-up discovered during review of the executor-to-commit columnar carrier cut.

Issue #1143, opened 2026-08-02
Stream certified entity row-group encoding to bound large UPDATE peak RSS


samuelstroschein:
## Problem

The certified `{path, value}` complete-replacement path builds row groups in `ROW_GROUP_MAX_ROWS` chunks, but `encode_certified_path_value_row_groups_impl` retains every unencoded Arrow `RecordBatch` in a `Vec` before `encode_row_group_set_preserving_batches` accumulates the complete encoded representation.

For 1M-row UPDATE this means the process simultaneously retains the full Arrow projection and the full encoded column set. Each group also creates transient `Vec<&str>`, `Vec<Option<&str>>`, and `Vec<String>` buffers before constructing the Arrow arrays.

## Profiling context

The direct certified projection cut reduces update allocations by ~27.6% (about 5.42 GB to 3.93 GB), but the remaining construction is not truly morselized and still contributes to peak RSS.

## Proposed direction

- Add a streaming/morsel row-group encoder that accepts and encodes one `RecordBatch` at a time while incrementally constructing the manifest and encoded columns.
- Build UTF-8 arrays directly with Arrow builders to avoid transient reference/string vectors.
- Preserve exact manifest, digest, statistics, and input-coordinate semantics.
- Benchmark 10K and 1M public SQL UPDATE on RocksDB and SlateDB; also measure update allocations and peak RSS.

This is intentionally a follow-up to the focused executor-to-commit columnar carrier optimization.

Issue #1142, opened 2026-08-02
`cargo clippy -p lix_engine --all-targets -- -D warnings` fails on unused StorageWriteSet::apply


samuelstroschein:
## Problem

On current `main` (including merge commit `0ce102abdda57f83c8ca48c3ac8682ca739ed814`), the strict all-targets Clippy invocation fails because `StorageWriteSet::apply` is unused in the test build:

```text
error: method `apply` is never used
  --> packages/engine/src/storage_adapter/write_set.rs:292:25
   |
292 |     pub(crate) async fn apply<StorageImpl>(
   |                         ^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
```

## Reproduction

```bash
cargo clippy -p lix_engine --all-targets -- -D warnings
```

## Expected

The strict all-targets Clippy invocation succeeds on `main`, either by removing the dead method if obsolete, exercising it, or narrowly documenting/expecting why it exists.

This was encountered while validating a separate CRUD performance change and is intentionally not included in that PR.

PR #1149, opened 2026-08-02
Reuse certified entity columnar batches

samuelstroschein:
## What changed

Carry the certified complete `{path, value}` replacement's typed identity/value projection from the SQL executor through transaction preparation and publish those already-encoded row groups at commit time.

This removes the redundant hot-path loop that serialized authoritative history JSON and then reparsed every row back into Arrow solely to build the current-generation analytical sidecar.

- retain canonical value offsets while the executor builds the authoritative snapshot arena
- encode the physically equivalent path/value/entity-PK row groups once
- preserve the derived carrier only while row order, schema, and complete-replacement proof remain exact
- fall back to authoritative generic semantics for partial batches, schema mutation/opt-out, and other certificate-invalidating shapes
- treat derived projection/physical-limit failures as sidecar fallback rather than rejecting valid SQL
- no changenote

JSON snapshots and commit deltas remain authoritative for history, diff, merge, and branch behavior. Durable row-group manifests, coordinates, and physical write counts are unchanged.

## Why

Profiling 1M public SQL UPDATE on #1136 showed two dominant Lix-only costs:

1. ~2.61s rebuilding Arrow from JSON at commit
2. ~2.28s in the tracked-root fence

This PR removes the first bottleneck. The root fence remains a separate stacked hard cut.

## Performance

Baseline is the immediately previous merged PR, #1136 (`0ce102abdda57f83c8ca48c3ac8682ca739ed814`). Same optimized benchmark fixture and public `executeBatch` bound UPDATE path.

| Rows | Adapter | #1136 | This PR | Change |
|---:|---|---:|---:|---:|
| 10K | RocksDB | 53.866 ms | 34.879 ms | **-35.2%** |
| 10K | SlateDB | 51.311 ms | 33.090 ms | **-35.5%** |
| 1M | RocksDB | 6.732 s | 4.147 s | **-38.4%** |
| 1M | SlateDB | 6.214 s | 3.547 s | **-42.9%** |

1M update-phase allocation accounting falls about 27.6% (RocksDB 5.424 GB → 3.925 GB; SlateDB 5.437 GB → 3.937 GB).

Physical accounting remains 31,216 puts, 2 deletes, and ~92.355 MB at 1M rows (298 puts, 2 deletes, ~1.750 MB at 10K).

Raw SQLite 1M UPDATE is 2.357s, so the remaining ratios are approximately 1.76× RocksDB and 1.51× SlateDB. This PR clears the stacked-PR threshold but does not claim the final 1.2× goal; the profiled tracked-root fence is next.

## Validation

- `cargo check -p lix_engine`
- `cargo clippy -p lix_engine --lib -- -D warnings`
- `RUST_MIN_STACK=16777216 cargo test -p lix_engine --lib`: 1,801 passed, 8 ignored
- public cross-adapter parity/OLAP suite: 8 passed
- exact direct-vs-generic row-group manifest/digest/coordinate equivalence, including JSON null
- complete 2,048-row replacement proves carrier reuse
- 1,024-of-2,048 partial replacement proves no unpublished sidecar is built
- derived projection failure proves authoritative UPDATE still commits via fallback
- staged schema amendment and columnar opt-out tests
- two independent adversarial reviews, including out-of-order/duplicate PK alignment and history/diff/merge/branch authority

Follow-up memory work is tracked in #1143 and #1144. The unrelated strict all-target lint issue on main is tracked in #1142.

### DIFF
--- a/packages/engine/src/session/execute.rs
+++ b/packages/engine/src/session/execute.rs
@@ -5710,7 +5710,8 @@ mod tests {
 
     #[tokio::test]
     async fn large_ordered_parameter_update_replaces_complete_packed_current_base() {
-        const ROW_COUNT: usize = 1_024;
+        const ROW_COUNT: usize = 2_048;
+        const PARTIAL_ROW_COUNT: usize = ROW_COUNT / 2;
         let session = open_session().await;
         let schema = serde_json::json!({
             "$schema": "https://json-schema.org/draft/2020-12/schema",
@@ -5750,6 +5751,7 @@ mod tests {
         session.execute_batch(&insert_statements).await.unwrap();
 
         crate::transaction::take_complete_replacement_packed_current_base_retirements();
+        crate::transaction::take_certified_entity_columnar_reuses();
         let update_sql =
             "UPDATE ordered_packed_update_probe SET value = lix_json($1) WHERE path = $2";
         for version in 1..=2 {
@@ -5771,6 +5773,11 @@ mod tests {
                 .sum::<u64>();
             assert_eq!(affected, ROW_COUNT as u64);
             assert_eq!(
+                crate::transaction::take_certified_entity_columnar_reuses(),
+                1,
+                "complete certified replacements should retain their executor-built columnar vectors"
+            );
+            assert_eq!(
                 crate::transaction::take_complete_replacement_packed_current_base_retirements(),
                 1,
                 "each complete certified replacement should swap one packed base reference"
@@ -5785,7 +5792,7 @@ mod tests {
 
         let rows = session
             .execute(
-                "SELECT path, value FROM ordered_packed_update_probe WHERE path IN ('0000', '1023') ORDER BY path",
+                "SELECT path, value FROM ordered_packed_update_probe WHERE path IN ('0000', '2047') ORDER BY path",
                 &[],
             )
             .await
@@ -5797,7 +5804,7 @@ mod tests {
         );
         assert_eq!(
             rows.rows()[1].get::<serde_json::Value>("value").unwrap(),
-            serde_json::json!("updated-2-1023")
+            serde_json::json!("updated-2-2047")
         );
         let working_diff = session
             .execute(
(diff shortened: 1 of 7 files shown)


samuelstroschein:
OLAP coordination ack. I will rebase phase instrumentation and packed scan cache work onto merge 196706c1.

Current generic analytical candidate reuses the existing repository-scoped bounded Arrow cache for pristine immutable row groups (content/state/projection keyed). Prototype SF0.2 RocksDB: Q8 -66.9%, Q11 -39.5%; point predicates/limits remain excluded. It does not change tracked-root writes or packed-base representation yet.

Packed profile: Q8 reads ~50.2 MB compressed into ~177.2 MB Arrow, with decode dominating storage materialization. For the tracked-root hard cut, please preserve stable immutable row-group content identity plus branch/head/revision/shadow identity sufficient for safe cache invalidation. Once the reviewed cache branch is public I will post its commit here and request your 10K/1M RocksDB+SlateDB CRUD/diff/merge regression.

samuelstroschein:
OLAP update: reviewed PR #1151 is open at https://github.com/opral/lix/pull/1151 (head 72eefb488). It broadens the existing bounded exact-projection Arrow cache to pristine immutable row groups; warm SF0.2 Q8 improves ~69–70% and Q11 ~36–37% on RocksDB+SlateDB. Exact PK/LIMIT paths are structurally excluded; 10k Rocks read_one guard is 9.201us/op (999/1000 point-cache hits). Sparse/moderate overlay parity passes. Cold decode remains unsolved.

Once #1151 is green/merged, please benchmark it with your 10k/1M CRUD + diff/merge suite against #1149/#1151 combined and flag any tracked-root hard-cut conflicts. The cache requires stable immutable RowGroupSetId/content and manifest column digests; branch/head/revision/shadow dimensions remain in the reconciled batch key.

samuelstroschein:
OLAP coordination update: #1151 is merged at 7ffcc952; next reviewed PR #1152 is open at https://github.com/opral/lix/pull/1152 (head 09ab614d5). It adds content-keyed immutable decoded-column reuse beneath #1151 and changes repository cache accounting to one shared 256MiB unique-ArrayRef identity/refcount budget. Transactions and exact-PK/LIMIT reads bypass; full Q1-Q22 correctness/counters match, worst normalized guard +3.39%. Q11 normalized ratio improves 12.9% Rocks/16.5% Slate.

Tracked-root layout constraint remains: immutable RowGroupSetId + canonical manifest/column content identity must remain stable; current cache key is row-group set + canonical manifest digest + group + column. Please include #1152 in the combined 10K/1M CRUD/diff/merge benchmark after it merges and flag conflicts in live_state/context/entity_batch/cache ownership.

samuelstroschein:
Coordination update: the user correctly identified x-lix-columnar as a physical-policy leak. I have an isolated removal branch based on 1c19577f that deletes the public schema key end-to-end and removes the CRUD benchmark exemption (10 files, +5/-220; not yet published pending serialized tests/review).

This means the next combined CRUD run must measure the normal universal layout policy rather than a schema-selected write-only path. Please do not preserve or reintroduce this flag in the tracked-root hard cut. The generic rule remains: every physically eligible registered entity schema may publish the immutable columnar base; query planning chooses paths from capabilities/cost, never schema workload intent. I will post the PR/commit after validation.

Issue #1199, opened 2026-08-04
whole-collection DELETE unit test deterministically overflows the test-thread stack


samuelstroschein:
## Problem

`sql2::exec::datafusion::tests::whole_entity_collection_delete_is_visible_inside_explicit_transaction` deterministically aborts with a stack overflow on current `main` (`9693d9fcf4c37ad43c0bc27be0b66682455a1eb3`).

## Reproduction

```sh
cargo test -p lix_engine --lib whole_entity_collection_delete_is_visible_inside_explicit_transaction -- --nocapture
```

Observed:

```text
thread 'sql2::exec::datafusion::tests::whole_entity_collection_delete_is_visible_inside_explicit_transaction' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The failure reproduces both in the full library suite and alone on exact `main`, so it is not introduced by the transaction-journal optimization branch. Diagnose the oversized stack frame or recursion and restore the default-stack test.

PR #1200, opened 2026-08-04
Seal transaction journal replacement parts eagerly

samuelstroschein:
## What changed

- eagerly seal bounded (≤16,384-row) typed transaction-journal chunks into unowned, immutable replacement-part bytes while their arenas are hot
- retain commit as the sole authority boundary: commit ID, lifecycle, replacement generation, directory, manifest, current state, and storage writes are still bound and published atomically
- consume a sealed canonical prefix directly and encode one contiguous suffix when a read barrier leaves an unsealed tail
- fail closed on adaptive >64 KiB part splits so residual rows cannot be finalized at a journal-chunk boundary
- hand compressed payloads into the storage write set through shared `Bytes` instead of copying them
- add a reproducible scalar-subset UPDATE profiling control

There is no storage-format compatibility path or changenote in this cut. Diff/history/merge/checkpoint/branch continue to consume the unchanged committed replacement-part and CommitStateManifest contracts.

## Why

The scalar SQL UPDATE path built the correct typed journal, but waited until commit to reconstruct keys and compress all replacement parts after the journal arenas had gone cold. For representative 10K complete replacements, sealing the existing journal buffers eagerly removes that commit spike without moving durable authority before commit.

Eager sealing is limited to collections whose complete generation is at most 16,384 rows. Large collections retain the canonical one-pass commit path.

## Performance

Exact immediate parent: merged #1198 (`60dcacb95`). Same checkout, release binary, feature set, fixture, and commands; medians of 21 samples.

| 10K scalar UPDATE | #1198 | this PR | change |
| --- | ---: | ---: | ---: |
| RocksDB | 32.552 ms | 25.813 ms | **-20.7%** |
| SlateDB | 32.044 ms | 25.290 ms | **-21.1%** |

Supplementary paired measurements taken immediately before the non-overlapping #1198 rebase, against merged #1197:

- 10K executeBatch: RocksDB -2.2%, SlateDB -1.0%
- 10K allocated bytes: +1.05%
- 1M peak RSS: +0.024%
- 1M scalar UPDATE: RocksDB +4.5%, SlateDB +6.6%, while remaining within the governing raw-SQLite target at 1.123× and 1.175×
- partial 4,096-of-10K scalar UPDATE: RocksDB +8.1%, SlateDB +2.0%

The primary UPDATE win clears the per-PR threshold. The 1M and partial-update regressions are explicitly accepted as an UPDATE-priority tradeoff and tracked in a follow-up issue.

## Correctness

- non-final adaptive-split regression test proves an oversized part cannot seal a non-canonical chunk boundary
- hybrid 8,193-row journal test covers sealed prefix + unsealed read tail, read-your-writes, atomic commit, and fresh commit ownership
- independent correctness and performance reviews approved the final cut
- `RUST_MIN_STACK=8388608 cargo test -p lix_engine --lib`: 1,885 passed, 8 ignored
- `cargo test -p lix_engine --test integration`: 440 passed, 2 ignored
- `cargo clippy -p lix_engine --all-features --all-targets -- -D warnings`
- `git diff --check`

The default-stack library suite currently aborts in an unrelated whole-collection DELETE test on both this branch and exact main; tracked in #1199.

### DIFF
--- a/packages/engine-benchmarks/benches/tracked_state_crud/sql_session.rs
+++ b/packages/engine-benchmarks/benches/tracked_state_crud/sql_session.rs
@@ -1008,6 +1008,10 @@ where
 
     #[expect(clippy::cast_possible_truncation)]
     async fn update_all(&self) -> usize {
+        let expected_rows = match &self.update_all_workload {
+            UpdateWorkload::Transaction(statements) => statements.len(),
+            UpdateWorkload::ExecuteBatch(_) => self.row_count,
+        };
         let affected = match &self.update_all_workload {
             UpdateWorkload::Transaction(statements) => {
                 Box::pin(execute_many_in_transaction(&self.session, statements)).await
@@ -1021,7 +1025,7 @@ where
                 .map(|result| result.rows_affected())
                 .sum(),
         };
-        assert_eq!(affected as usize, self.row_count);
+        assert_eq!(affected as usize, expected_rows);
         affected as usize
     }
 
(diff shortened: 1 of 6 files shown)


samuelstroschein:
Follow-up performance work for the explicitly disclosed 1M and partial scalar UPDATE regressions is tracked in #1201.

PR #1228, opened 2026-08-07
Profile pull-based SQL result streaming

samuelstroschein:
## Summary

- preserve the current eager `ExecuteResult` API while adding a benchmark-only pull-based DataFusion result primitive
- bind live batch results and the internal batch-to-row cursor to the read-session lifetime so borrowed storage reads cannot escape
- compare eager rows, collected Arrow batches, live batches, and count-only execution with distinct scanned/consumed/materialized/retained counters
- add a fresh-process randomized RSS/checksum matrix, including fixture-only and collected-all controls
- add synthetic cursor-drop coverage and a real SessionContext/DataFusion early-drop test

No JS, N-API, WASM, worker, remote, CLI, filesystem, storage, or integration-suite API migration is included.

## Validation

- `cargo check -p lix --no-default-features --features storage-benches --bench profile_sql_result_streaming`
- `cargo test -p lix --no-default-features --features storage-benches live_profile_early_drop_stops_before_later_datafusion_partition --lib`
- `cargo test -p lix --no-default-features batch_cursor --lib`
- `cargo test -p lix --no-default-features --features storage-benches sql_profile::tests --lib`
- reduced fresh-process eight-scenario matrix at 1,024 rows

The reduced matrix showed the intended causal separation for a 100-row early stop: eager and collected modes scanned all 1,024 rows, while live mode scanned one 128-row batch. Checksums over all projected values passed. RSS is reported against a fixture-only process baseline and remains a process high-water measurement.

### DIFF
--- a/packages/lix/Cargo.toml
+++ b/packages/lix/Cargo.toml
@@ -52,6 +52,12 @@ path = "tests/semantic_merge.rs"
 name = "send_futures"
 path = "tests/send_futures.rs"
 
+[[bench]]
+name = "profile_sql_result_streaming"
+path = "benches/profile_sql_result_streaming.rs"
+harness = false
+required-features = ["storage-benches"]
+
 [[test]]
 name = "plugin_arena_scorecard"
 path = "tests/plugin_arena_scorecard.rs"
--- a/packages/lix/benches/profile_sql_result_streaming.md
+++ b/packages/lix/benches/profile_sql_result_streaming.md
@@ -1,0 +1,105 @@
+# SQL result streaming profile
+
+This benchmark compares the current eager public result path with two internal
+prototypes. It does not change `ExecuteResult` or expose streaming publicly.
+Every mode uses the same fixture and SQL statement, and the profile scope
+includes snapshot acquisition, execution, and result consumption.
+
+The fixture query is an unordered `UNION ALL` over eight distinct registered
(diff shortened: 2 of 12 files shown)


samuelstroschein:
CI note: the original Clippy failure was inherited from `main` at `fdd062078` (the base run failed with the same 19 library and 5 test diagnostics). Commit `f81304c64` contains a separate mechanical cleanup for those existing tree-sweep/test lints. The exact CI command now passes locally:

```sh
```

samuelstroschein:
Cargo Test is also failing on the base `main` commit, independently of this PR. Base run [31130763554](https://github.com/opral/lix/actions/runs/31130763554) and PR run [31137218471](https://github.com/opral/lix/actions/runs/31137218471) both fail the same six `tracked_state_rebuild` cases with the same deterministic seeds and missing/mismatched tree-root errors:

- three `checkpoint_gc` cases
- `corruption_fuzz` seed `0x1`, transaction 15
- `merge_fuzz` seed `0x0`, step 4
- `filesystem_fuzz` seed `0x1`, step 39

The streaming diff does not touch those execution paths. Clippy is green after the separate baseline lint cleanup; resolving Cargo Test requires an upstream tracked-state rebuild/GC correctness fix rather than suppressing tests in this PR.

PR #1270, opened 2026-08-11
experiment: give untracked rows terminal change ids

samuelstroschein:
## Experiment

This is a standalone experiment against `origin/main` (`dab4f5c686c2a65a4d422a2ffb696180d5eea2c4`). It is unrelated to the ForkTree draft.

It tests whether durable untracked current-state rows should participate in the same logical `lix_change` identity stream as tracked rows:

- every durable untracked row owns a non-nil `ChangeId` and no `CommitId`;
- the current change is visible through `lix_change` but is not a commit member;
- replacing a row atomically stages the new standalone change and eagerly retires the previous one;
- deleting a row retires its final change without creating tombstone history;
- GC roots an untracked payload through its standalone change rather than through a separate current-payload authority.

The resulting changelog cardinality is bounded by live untracked rows, not update count.

## Correctness

- `cargo check -p lix --tests --all-features`
- full `lix` lib suite: **2056 passed, 0 failed, 8 ignored**
- untracked SQL lifecycle: **14 passed, 0 failed**
- hot-state tests: **47 passed, 0 failed**
- insert/replace/delete tests prove one live change, old-ID retirement, no commit member, and no tracked working-diff entry.

## Paired benchmark

The existing byte-identical `untracked_state_crud` public SQL harness was compiled from both the parent and candidate. RocksDB and SlateDB each ran 3 warmups plus 20 measured 1K-row samples; setup was excluded. Harness SHA-256: `205ce053c79d3ed73edbeac07421b41d65498d15c4d420333e1e6aed9b944de4`.

| Backend | Operation | main median / p95 | candidate median / p95 | latency delta | written bytes main → candidate |
| RocksDB | insert | 14.938 / 19.307 ms | 15.456 / 16.150 ms | +3.5% | 417,089 → 673,643 (+61.5%) |
| RocksDB | update | 47.665 / 47.837 ms | 47.511 / 47.786 ms | -0.3% | 1,541,552 → 1,692,796 (+9.8%) |
| RocksDB | delete | 4.814 / 4.986 ms | 5.188 / 5.294 ms | +7.8% | 222 → 222, plus 1,000 deletes |
| SlateDB | insert | 20.950 / 23.258 ms | 21.843 / 22.921 ms | +4.3% | 417,089 → 673,643 (+61.5%) |
| SlateDB | update | 47.782 / 48.460 ms | 47.704 / 48.997 ms | -0.2% | 1,541,552 → 1,692,796 (+9.8%) |
| SlateDB | delete | 6.408 / 6.611 ms | 6.235 / 7.162 ms | -2.7% | 222 → 222, plus 1,000 deletes |

Summary TSV SHA-256: `bcc1c19a61c8d09ddd9253b2c4a77207ffc4685196d772c0283f84cbb78fc063`.

## Data-driven verdict

The unified identity model is operationally viable: update latency is neutral and insert/delete regressions are small. It is not free, however. Persisting one terminal changelog object per live untracked row increases initial write bytes by ~61.5%; replacement adds ~9.8% write bytes and one delete per row.

This therefore supports the change only if the simpler single logical change stream is worth that storage/write amplification. It is not a performance optimization.

## Deliberate draft limitations

- Existing hot-row v8 data may contain nil IDs for untracked rows. This candidate fails closed on those rows; a repository migration must be designed before merge.
- The SQL `change_id` field remains nullable because pending tracked writes intentionally hide provisional IDs until commit assigns canonical ordering. All **durable** tracked and untracked rows have an ID. Making the schema globally non-null requires moving canonical tracked-ID assignment earlier and is a separate semantic change.
- Old standalone changes are retired eagerly in the replacement publication rather than by a later GC pass. This gives the requested bounded stream while avoiding a period with two public current changes.

## Provenance

- head: `0ec87b77bb85214f9c9eee168d3dcf65f92c122a`
- tree: `252ef39a3e7a2cf244663ffe7815ea08dd9751fc`
- parent: `dab4f5c686c2a65a4d422a2ffb696180d5eea2c4`
- parent..head full-index SHA-256: `8f66e485bfe14b76b226891ab170cf239e10411dc06bab8aa83cd1de520cac99`
- stable patch-id: `f5e4e4be8bf3104f916e8d76783f544030ee7e3c`

### DIFF
--- a/packages/lix/src/changelog/context.rs
+++ b/packages/lix/src/changelog/context.rs
@@ -28,7 +28,7 @@ use crate::changelog::{
     ChangeId, ChangeLoadBatch, ChangeLoadRequest, ChangeRecord, ChangeScanBatch, ChangeScanRequest,
     ChangelogAppend, ChangelogReader, ChangelogWriter, CommitId, CommitLoadBatch,
     CommitLoadRequest, CommitRecord, CommitScanBatch, CommitScanRequest,
-    TransactionChangelogAppend,
+    TransactionChangeRecordRef, TransactionChangelogAppend,
 };
 use crate::json_store::JsonSlotRef;
 use crate::storage_adapter::Storage;
@@ -48,6 +48,23 @@ pub(crate) struct ChangelogContext;
 impl ChangelogContext {
     pub(crate) fn new() -> Self {
         Self
+    }
+
+    /// Stages one normalized standalone change in the caller's existing
+    /// publication write set. This is the no-second-read terminal path for
+    /// engine-owned untracked current state.
+    pub(crate) fn stage_terminal_standalone_change(
+        &self,
+        writes: &mut StorageWriteSet,
+        change: TransactionChangeRecordRef<'_>,
+    ) -> Result<(), LixError> {
+        let mut batch =
+            EncodedChangelogBatch::with_capacity(1, 16, transaction_change_value_capacity(&change));
+        batch.try_put(change.change_id.as_uuid().as_bytes(), |bytes| {
+            append_transaction_change_record(bytes, &change)
+        })?;
+        batch.stage(writes, CHANGE_SPACE);
+        Ok(())
     }
 
     pub(crate) fn reader<S>(&self, store: S) -> ChangelogStoreReader<S>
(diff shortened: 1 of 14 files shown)


samuelstroschein:
Superseded by #1346, which landed the same capability with a different cut.

#1270 made the untracked change visible through `lix_change`, measured at **+61.5% written bytes on insert** and +9.8% on update (RocksDB, 1k rows). #1346 mints the id but writes **no changelog record** — `commit.rs`'s "untracked state is history-free and must not enter the changelog" assertion stays true and stays asserted — and measured **0.00% byte delta**, reproduced exactly across all 9 reps on both RocksDB and SlateDB, with timing inside a null-control band of 25.6%/47.9%.

This PR's measurement is what located the cost in changelog participation rather than the 16-byte id, and that attribution is why #1346 took the shape it did. Closing as superseded, not as wrong.

PR #1304, opened 2026-08-11
perf(binary-cas): anchored content-defined chunking [measured, recommend closing]

samuelstroschein:
> **Recommendation: close this PR.** The gate that would have made the trade
> acceptable — CDC on rewrites, fixed chunking on initial ingest — is not just
> awkward to implement, it is **measurably worse than doing nothing**. Numbers
> below. The measurement stands as the record; the durable parts of this work
> already landed in #1303 and #1309.

## Why the rewrite gate cannot work

Deduplication requires v1 and v2 of a file to be chunked **by the same rule**.
That is the entire mechanism: chunks match because boundaries land in the same
places. Gating the chunker on write context guarantees the opposite — the initial
write gets fixed 1 MiB boundaries, the rewrite gets content-defined ones, and a
content-defined boundary essentially never lands on a 1 MiB multiple.

Measured with `chunking_oracle` over the same real-file corpus (branch
`exp/chunking-oracle-gate`, scenario `mixed/fixed-v1_cdc-v2`):

| policy | stored bytes for v2 | shared |
|---|---:|---:|
| `fixed/1m` (today) | 529,835,742 | 0.451 |
| `cdc/1m@16m` (this PR) | 204,935,691 | **0.788** |
| **`mixed/fixed-v1_cdc-v2` (the gate)** | **964,697,149** | **0.001** |

The gate stores **+82% more bytes than shipping nothing at all**. Every shape
drops to 0.000 — including the ones fixed chunking already handled well, such as
`localized overwrite` (0.909 → 0.000) and `image_bmp/crop_8px` (0.975 → 0.000).

To be precise about what that row models: it is the **first** rewrite of a blob.
Under the gate, second and later rewrites would be CDC-vs-CDC and would dedup
normally. So the gate's real shape is "every blob pays one full rewrite the first
time it is edited, then behaves like CDC" — while still paying CDC's CPU on every
rewrite, which is where the cost is largest (`full_rewrite` +64%,
`localized_4k_update` +109%). It removes CPU only from the initial write (+55%),
which is exactly the lane you measured as gaining nothing.

There is a second, independent blocker: blob identity is *derived from the chunk
layout* (`BlobId::from_chunks`, keyed `"lix binary cas ... manifest"`). A
context-dependent layout gives identical bytes two different `BlobId`s, which
breaks cross-file dedup, the `prepare_blob_write` debug assertion comparing the
precomputed hash against `BlobId::from_content`, and the verification at
`filesystem/path_index.rs:633`. But the sharing table settles it without needing
that argument.

## The cheaper-CDC alternative, also measured, also insufficient

FastCDC skips `min` bytes without hashing at the start of each chunk, so raising
the minimum is directly less CPU:

| minimum | boundary search | corpus shared |
|---|---:|---:|
| avg/4 (this PR) | 4,474 MiB/s | 0.788 |
| avg/2 | 5,860 MiB/s | 0.785 |
| avg | **11,026 MiB/s** | 0.771 |

2.46x cheaper for 1.7 points of sharing — a good trade in isolation. But it only
takes the measured `initial_write` cost from +55% to roughly +23% (10 MiB: 4.04 ms
base, 2.24 ms of CDC at avg/4, 0.91 ms at avg). Still well outside the ±6%
null-control floor, still paid on a lane that your ingest run shows gains
**1.9 KB out of 675 MB**. It does not change the verdict.

## What the measurement established

Worth keeping in the record even though the change should not land:

- Fixed 1 MiB offsets share **exactly zero** on every insert and delete, and on a
  metadata-only MP4 remux that leaves the H.264 payload untouched. 21 of 47
  real-file case/shape pairs.
- Content-defined boundaries recover that: corpus stored bytes −66.5%, sharing
  0.459 → 0.818, and −57.6% on the rewrite lane specifically
  (`localized_4k_update` at 10 MiB, 12,596,217 → 5,338,799 bytes).
- That win is **confined to rewrites of an existing blob**. Initial ingest of
  unique footage has nothing to find, and pays +55%.
- Anchoring CDC at `FILE_UPLOAD_PART_BYTES` does preserve #1084's bounded upload
  memory and the four-part out-of-order window — the objection that removed
  FastCDC originally is answerable. It is the CPU-versus-benefit distribution,
  not memory, that sinks it.
- Under CDC a 512 MiB ingest produces 447 chunk rows rather than 512, i.e. ~13%
  *fewer* manifest rows. The qualification assertion is updated to a bound rather
  than an equality in this branch, since fewer, larger chunks only improve the
  projection.

## Branch state

Rebased onto `0839fdea7`. `cargo test -p lix` passes. Kept green and truthful so
it can be revisited if a workload appears whose dominant cost is repeated
rewriting of large binaries — that is the only profile this pays for.

### DIFF
--- a/.changenotes/content-defined-binary-chunking.md
+++ b/.changenotes/content-defined-binary-chunking.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+Binary files now reuse storage across edits that shift bytes, such as rewriting an MP4's metadata, trimming a clip, or splicing audio.
+
+The binary content store previously cut chunks at fixed 1 MiB offsets, so inserting or removing bytes anywhere in a file renamed every chunk after that point and the whole file was stored again. Chunk boundaries are now chosen from the content itself, so the boundaries resynchronise just past the edit and the rest of the file keeps the storage it already had. Chunk sizes still average 1 MiB, so file layout, seek behaviour, and row counts are unchanged.
--- a/packages/engine-benchmarks/tests/movie_workspace_qualification.rs
+++ b/packages/engine-benchmarks/tests/movie_workspace_qualification.rs
@@ -506,8 +506,20 @@ where
         assert_eq!(first_late, 0, "first 100 Mbit/s stream fell behind");
         assert_eq!(second_late, 0, "second 100 Mbit/s stream fell behind");
         assert_eq!(structural.temporary_manifest_leaf_rows, 32);
-        assert_eq!(structural.legacy_equivalent_chunk_rows, 512);
-        assert_eq!(row_reduction, 16.0);
+        // Chunk boundaries are content-defined, so a 512 MiB ingest no longer
+        // yields exactly 512 one-megabyte chunks. What the projection depends on
+        // is that a part still summarizes many chunks into one leaf, so assert
+        // the reduction is at least the fixed-layout ratio rather than equal to
+        // it: fewer, larger chunks only make it better.
+        assert!(
+            structural.legacy_equivalent_chunk_rows <= 512,
+            "content-defined chunking must not add chunk rows, got {}",
+            structural.legacy_equivalent_chunk_rows
+        );
+        assert!(
+            row_reduction >= 13.0,
+            "each upload part must still summarize many chunk receipts, got {row_reduction}"
+        );
         assert_eq!(projected_chunk_rows / projected_leaf_rows, 16);
         assert!(structural.chunk_payload_hash_bytes >= INGEST_BYTES as u64);
     }
--- a/packages/lix/src/binary_cas/chunking.rs
+++ b/packages/lix/src/binary_cas/chunking.rs
@@ -1,31 +1,65 @@
 //! The single authority for binary-CAS chunk boundaries.
 //!
 //! Boundaries are decided exactly once per payload, in
-//! [`crate::binary_cas::BlobPayload::from_bytes`], and travel with the payload
-//! as chunk receipts. Staging reconstructs the ranges from those receipt sizes
-//! rather than asking this module again, so a write never searches for the same
-//! boundaries twice.
+//! [`crate::binary_cas::BlobPayload::from_bytes`] for a whole buffer or in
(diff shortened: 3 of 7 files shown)


samuelstroschein:
Closing on the measurement, with the record preserved in the PR body and in the oracle landed as #1321.

**The decision:** CDC buys −66.5% stored bytes, but only on *repeated edits to existing large binaries*. On the qualified media workload — 675 MB of ingested footage — it saves **1.9 KB** and costs **+55% initial_write / +64% full_rewrite** CPU, all far outside the ±6% null-control floor. Under this project's OLTP > physical-bytes ranking that is the wrong trade.

**The gate I proposed (CDC on rewrites, fixed on initial writes) is disproven, and worse than shipping nothing.** Deduplication requires v1 and v2 to be chunked by the same rule — that *is* the mechanism. Gating on write context guarantees v1 gets fixed 1 MiB boundaries and v2 gets content-defined ones, which essentially never coincide:

| policy | stored bytes for v2 | shared |
|---|---:|---:|
| fixed/1m (today) | 529,835,742 | 0.451 |
| cdc/1m@16m (this PR) | 204,935,691 | 0.788 |
| mixed fixed-v1 / cdc-v2 (the gate) | **964,697,149** | **0.001** |

+82% versus doing nothing, with every shape collapsing to 0.000 — including ones fixed chunking already handled well. A second independent blocker: `BlobId` is derived from the chunk layout, so context-dependent chunking would give identical bytes two different blob ids, breaking cross-file dedup.

**What survives and has already landed** (#1309): the single-pass boundary decision, `chunking.rs` as the sole authority for boundaries, removal of the dead `BinaryCasChunking` plumbing, and the `fastcdc_*` names no longer claiming a property the code never had — the original complaint that started this work.

**What is preserved for a future attempt** (#1321): asymmetric-scenario and minimum-size oracles. Notable result — raising FastCDC's minimum from avg/4 to avg makes boundary search **2.46× cheaper** (4,474 → 11,026 MiB/s) for 1.7 points of sharing. Not enough to change this verdict, but the right starting point if the workload ever shifts toward repeated large-binary editing.

PR #1310, opened 2026-08-11
test(movie-workspace): retry saves the engine asks to retry

samuelstroschein:
## What this is

`movie_workspace_qualification` fails on `claude-3-optimizations` today. Both
`slatedb_movie_workspace_interference` and
`slatedb_movie_workspace_latency_simulation` panic at
`movie_workspace_qualification.rs:504` with:

```
save project file: LixError { code: "LIX_TRANSACTION_CONFLICT",
  message: "transaction snapshot is stale because tracked state changed before commit",
  hint: Some("Retry the transaction against the latest committed state.") }
```

The test does not retry, so it takes the engine's "please retry" as a hard
failure. This commit retries the save up to 32 times, counts the retries, and
reports them as `save_conflict_retries` in the `movie_workspace,...` line.
Retrying is what a real editor does and the retry time belongs inside the save
latency the qualification gates on, so `save_p95` stays meaningful.

## It is not sufficient on its own

With the save site fixed, the run gets further and then fails at `:450`:

```
upload movie part: LixError { code: "LIX_TRANSACTION_CONFLICT", ... }
```

That is the **four-part out-of-order upload window** — the concurrency guarantee
PR #1084 was built around — conflicting against the concurrent project saves. I
deliberately did not add a retry there: the point of that window is concurrent
completion, and wrapping it in retries would hide the behaviour rather than
measure it.

So this PR is offered as a partial unblock and a bug report. The qualification
cannot currently produce a number on any branch, which also blocks the
remote-latency ingest profile needed to evaluate #1304.

Measured on `ryzen-9950x-IV`. Both failures reproduce identically on base
`79735d753` and on unrelated branches, so neither is caused by any binary-CAS
change under review.

### DIFF
--- a/packages/engine-benchmarks/tests/movie_workspace_qualification.rs
+++ b/packages/engine-benchmarks/tests/movie_workspace_qualification.rs
@@ -32,6 +32,8 @@ const STREAM_PERIOD: Duration = Duration::from_millis(80);
 const PLAYBACK_READ_AHEAD: Duration = Duration::from_secs(1);
 const STREAM_SAMPLES: usize = 40;
 const SAVE_SAMPLES: usize = 40;
+const SAVE_CONFLICT_RETRIES: usize = 32;
+static SAVE_CONFLICT_RETRY_COUNT: AtomicU64 = AtomicU64::new(0);
 const HISTORY_COMMITS: usize = 5_000;
 const TIB: u64 = 1024 * 1024 * 1024 * 1024;
 const GIB: u64 = 1024 * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


samuelstroschein:
Superseded by #1314, which fixed the root cause rather than retrying around it.

This PR added a retry at the `:504` project-save site because the engine asked for one. #1314 showed the engine should never have asked: the binary-CAS mutation epoch was a single row that **every** publisher both rotated and held a `KeyValueEquals` CAS on, making ordinary commits, upload parts and the CAS sweep all mutually exclusive with each other. Splitting it into an asymmetric fence (reclamation key CAS'd by sweeps, publication key written without a self-precondition) removed publisher-to-publisher exclusion while preserving publisher-to-sweep exclusion in both directions.

With that merged, `rocksdb_movie_workspace_interference` passes and both conflict sites are gone on SlateDB.

Closing deliberately rather than merging: keeping a retry in a **qualification test** would mask exactly this class of regression if it returned. The qualification should assert that concurrent unrelated writers do not conflict, which is now the engine's actual contract.

The branch is left pushed as a record of the intermediate diagnosis.

PR #1326, opened 2026-08-11
perf(json-store): reject zstd level 3 — 3% slower on update_all_rows despite 3-9% fewer bytes

samuelstroschein:
## Verdict: do not merge. This is a measured negative result.

This branch raises the JSON store's zstd level from 1 to 3. It was expected to be
**strictly dominant** — smaller *and* faster on every axis — and on that basis worth
landing despite being far below the 10% acceptance gate.

**It is not dominant on the current head.** It carries a real, reproducible **+3.0% /
+3.3% regression on `update_all_rows`** on both storage backends, confirmed at 9 reps with
completely non-overlapping raw ranges and against a null control that puts the same lane
at +0.2% / +0.6%. Under the trade-off ranking (**OLTP > version control > physical bytes >
OLAP**), buying 2.8-8.8% of physical bytes with a 3% OLTP write regression is the wrong
direction, and there is no >10% win anywhere in the end-to-end numbers to pay for it.

I am opening this as the documented negative result, with the harness improvements that
made the result visible. The measurement code is worth keeping; the level change is not.

Earlier numbers on this branch (taken before several merges) reported the encode side as
"lower CPU" and decode as "unchanged". Re-measured on `bdd8d3467`, **both of those were
wrong**, in opposite directions — see below.

## What changed

- `packages/lix/src/json_store/compression.rs` — `JSON_ZSTD_LEVEL = 3` (was an inline `1`).
- `packages/engine-benchmarks/examples/expu_json_delta.rs` — `LIX_EXPU_ENCODE_LEVELS` /
  `LIX_EXPU_ENCODE_DELTA` so the same level can be timed in more than one slot, giving the
  run its own null control and rotated arm order; a decode-timing block; and a fix to
  `today_bytes`, which indexed slot 0 of the level list and so silently stopped meaning
  "what the store actually writes" the moment the store's level moved off the first slot.

Nothing was removed. No space, writer, or code path was cut.

## Scope: the level-1 call sites outside the JSON store are deliberately untouched

`packages/lix/src/compression.rs` still exposes `compress_zstd_level_1` and
`ZstdLevel1Compressor`, and their callers still run at level 1: the tracked-state tree
chunks and data/replacement parts (`tracked_state/{scoped_range,storage,replacement_part,
current_state_data_part}.rs`), the mutation journal (`transaction/{context,staging}.rs`),
`columnar_row_group.rs`, `plugin/incremental.rs`, and `wasm/component.rs`.

**I did not touch them, and that was the correct scope** — a judgment the result has now
strongly vindicated:

1. They are a different payload population. The JSON store compresses documents above a
   512-byte floor; those sites emit 256 KiB certified-plugin pages, up to 16 MiB tree
   nodes, and up to 256 MiB column chunks.
2. **I had no measurement for them** — and the whole point of this PR's finding is that a
   level change that looks free on one payload population is *not* free on another. Having
   just been burned by exactly that on the JSON store, propagating the same unmeasured
   change to five more call sites on the commit path would have been indefensible.

The constant is named `JSON_ZSTD_LEVEL` and lives in the JSON store, so the two remain
independently tunable rather than coupled through one shared helper. That part is worth
keeping regardless.

## Layout invariants checklist

A compression-level constant touches none of them. Stating that plainly rather than
manufacturing relevance:

1. **Single authority.** No. No second writer, no second materialization, no parallel
   index. The JSON store was and remains the single authority for its payloads; only the
   parameter it serializes them with changed.
2. **Commit canonicality.** Unchanged. No commit record, parent link, or state root is
   read or written differently; a branch ref still points at exactly one canonical commit.
3. **Derived views are caches.** Unchanged. Nothing derived is added, removed, re-tagged,
   or promoted to truth.
4. **Atomic publication.** Unchanged. The frame is produced during staging and published
   in the same write set that already published it; the publication boundary does not move.
5. **GC from refs.** Unchanged. Reachability still starts at refs and walks
   commits → state nodes → blobs; it does not consult payload encoding, and no retention
   metadata was added.
6. **SQL reads canonical records.** Unchanged. SQL reads the same canonical
   `json_store.json` rows. A zstd frame records its own parameters, so no reader knows or
   needs to know which level wrote it — which is why this needed no migration, no fallback
   reader, and no dual writer.

## Measurements

Machine `ryzen-9950x-III` (32 cores, 186 GB), sole occupant, warm build, load settled
below 1.0 before every timing window. Base = `claude-3-optimizations` @ `bdd8d3467`.
Corpora are seeded through the real `SessionContext` SQL commit path.

### 1. The regression (the decisive result)

`cargo bench -p lix_benchmarks --features storage-benches,slatedb --bench tracked_state_crud`,
filtered to the write lanes. **9 reps per arm, interleaved and rotated** so each arm takes
each position an equal number of times.

| lane | base median (ms) | cand median (ms) | delta | ranges overlap? |
|---|---:|---:|---:|:--|
| `lix_rocksdb/.../update_all_rows/10k` | 7.7548 | 7.9860 | **+2.98%** | **no** |
| `lix_slatedb/.../update_all_rows/10k` | 7.7549 | 8.0111 | **+3.30%** | **no** |
| `lix_rocksdb/.../insert_all_rows/10k` | 25.7200 | 25.8720 | +0.59% | yes |
| `lix_slatedb/.../insert_all_rows/10k` | 25.5110 | 25.5930 | +0.32% | yes |

Raw per-rep, `update_all_rows` (ms):

```
rocksdb base  7.6670 7.7331 7.7421 7.7479 7.7548 7.7557 7.7598 7.7729 7.8110
rocksdb cand  7.9012 7.9457 7.9613 7.9620 7.9860 8.0035 8.0099 8.0158 8.0316
slatedb base  7.6849 7.7329 7.7463 7.7504 7.7549 7.7561 7.7849 7.8053 7.8131
slatedb cand  7.9763 7.9819 7.9931 7.9950 8.0111 8.0149 8.0205 8.0206 8.0511
```

The base and candidate ranges do not touch on either backend. This is not a resolution
argument.

### 2. Null control for that bench

The obvious objection is that two worktrees produce different binaries. So I put the
candidate worktree on **the same commit as base** (`bdd8d3467`), rebuilt it in place, and
re-ran the identical 6-rep rotated A/B. Two byte-identical builds, two different worktrees
and target dirs:

| lane | base median | "cand" median (identical source) | delta |
|---|---:|---:|---:|
| `lix_rocksdb/.../update_all_rows/10k` | 7.7613 | 7.7777 | **+0.21%** |
| `lix_slatedb/.../update_all_rows/10k` | 7.7696 | 7.8125 | **+0.55%** |
| `lix_rocksdb/.../insert_all_rows/10k` | 25.6400 | 25.5960 | -0.17% |
| `lix_slatedb/.../insert_all_rows/10k` | 25.5710 | 25.6390 | +0.27% |
| `lix_rocksdb/.../read_all_rows_consumed/10k` | 3.7666 | 3.7309 | -0.95% |
| `lix_slatedb/.../read_all_rows_consumed/10k` | 3.8065 | 3.8051 | -0.04% |
| `lix_rocksdb/.../delete_all_rows/10k` | 0.5866 | 0.5858 | -0.14% |
| `lix_slatedb/.../delete_all_rows/10k` | 0.7364 | 0.7288 | -1.03% |

The worktree artifact on `update_all_rows` is **+0.2% / +0.6%**. The measured regression is
**5-6x that**. The null control does not explain it.

Cleanest form of the same comparison: in the **same worktree, same target directory**, the
`update_all_rows` median was **7.778 / 7.813 ms at level 1** and **7.994 / 8.011 ms at
level 3**. Only the constant differs.

### 3. The other guard lanes (no regression)

From the 6-rep rotated guard run. Everything except the write lanes sits inside its own
spread and inside the null control:

| lane | delta | base spread | cand spread | null-control delta |
|---|---:|---:|---:|---:|
| `rocksdb/read_all_rows_consumed/10k` | +0.28% | 1.72% | 1.30% | -0.95% |
| `slatedb/read_all_rows_consumed/10k` | +0.37% | 2.04% | 1.50% | -0.04% |
| `rocksdb/read_all_rows/10k` | +0.53% | 2.56% | 1.72% | -1.37% |
| `slatedb/read_all_rows/10k` | +0.17% | 2.19% | 1.93% | -0.24% |
| `rocksdb/delete_all_rows/10k` | +2.19% | 3.18% | 4.25% | -0.14% |
| `slatedb/delete_all_rows/10k` | +1.83% | 3.98% | 2.66% | -1.03% |

`delete_all_rows` looks like +2%, but it is a 0.6-0.7 ms lane whose own null-control spread
is 2.7-7.5%, and its write accounting is byte-identical between arms (183,236 bytes both) —
it writes no JSON payloads at all, so it cannot be affected. That is noise, not a signal.

`read_all_rows_consumed` — the primary read metric — is **flat**. The -23% decode win in
§6 does not reach the end-to-end read path in this workload, because only 22 JSON rows are
involved.

### 4. Settled bytes and write accounting (deterministic, 1 rep per arm)

`LIX_TRACKED_STATE_CRUD_ACCOUNTING=1`, 10k rows. This is the proof that the change is
scoped where it claims to be — and it is also what makes the regression attributable:

| space | rows | key bytes | value bytes @ L1 | value bytes @ L3 | delta |
| `json_store.json` | 22 | 792 | 151,126 | 148,322 | **-1.86%** |

**Every other space is byte-identical between the arms** — `tracked_state.tree_chunk`,
`tracked_state.commit_delta_segment.v6`, `live_state.hot_row.v21`,
`tracked_state.change_locator.v2`, `branch.head_control.v10`,
`tracked_state.commit_state_manifest.v7`, `tracked_state.commit_mutation_catalog.v1`, and
the rest, all unchanged to the byte. Staged puts, staged deletes, touched spaces, storage
calls, and put/delete batch counts are identical in every lane.

| operation | written bytes @ L1 | written bytes @ L3 | delta |
|---|---:|---:|---:|
| `transaction/insert_all` (10k) | 562,201 | 559,446 | -0.49% |
| `transaction/update_all` (10k) | 622,256 | 619,533 | -0.44% |
| `transaction/delete_all` (10k) | 183,236 | 183,236 | 0.00% |

So the candidate writes **fewer** bytes and touches nothing else, and is still 3% slower on
the update lane. The extra time is CPU spent inside the JSON compressor.

### 5. Bytes on synthetic corpora — deterministic, 1 rep per corpus per arm

Byte counts reproduce exactly, so these are measured once and the machine time spent on the
guards instead. Both arms are read back with the **same** binary, so the reader is a
constant. All seven corpora the harness supports, including the ones that flatter the
change least:

| corpus | rows | raw bytes | stored @ L1 | stored @ L3 | delta |
| doc_small | 10,403 | 17,080,852 | 6,522,480 | 6,340,603 | -2.79% |
| kv_rewrite | 22,003 | 92,360,052 | 23,677,813 | 21,994,177 | -7.11% |
| kv | 22,003 | 92,360,052 | 23,704,031 | 22,001,221 | -7.18% |
| doc_edits | 10,403 | 113,988,052 | 25,898,562 | 24,006,013 | -7.31% |
| doc_header_only | 10,403 | 113,988,052 | 25,696,762 | 23,819,631 | -7.31% |
| doc_scattered | 10,403 | 113,988,052 | 25,898,562 | 24,006,013 | -7.31% |
| doc_large | 10,403 | 932,676,052 | 183,353,231 | 167,254,982 | -8.78% |

Two caveats, both against my own case:

- **`doc_scattered` is not an independent data point.** Its settled store is byte-identical
  to `doc_edits` — the corpora differ only in *which* blocks each edit touches, and the
  JSON store holds only settled payloads. Six independent corpora, not seven.
- `raw_bytes` is identical between arms in every row, which is the check that the logical
  payload stream really is the same and only the encoding moved.

SlateDB directory bytes for the same stores, reported for completeness and **not** as a
headline — `storage.flush()` runs before this is sampled, so it includes un-compacted LSM
state and swings much wider than the logical figure: doc_small -1.13%, doc_scattered
-3.40%, doc_large -8.71%, kv_rewrite -8.83%, kv -8.92%, doc_header_only -10.24%,
doc_edits -11.84%.

### 6. Encode and decode on those same corpora — where the prediction came from

`LIX_EXPU_ENCODE_LEVELS=1,3,3,1` times four arms per rep in **one process on one payload
stream**, so the engine build is not a variable. Level 1 occupies slots 0 and 3, level 3
occupies slots 1 and 2: slots running byte-identical code give the null control, and each
level sitting both early and late cancels position drift. 9 reps, ns per payload, median:

| corpus | payload p50 | L1 slot0 | L1 slot3 | L3 slot1 | L3 slot2 | L3 vs L1 | null spread |
|---|---:|---:|---:|---:|---:|---:|---:|
| kv | 4,163 B | 8,539 | 8,541 | 8,317 | 8,335 | -2.51% | 0.02% / 0.22% |
| doc_edits | 10,931 B | 17,408 | 17,428 | 16,759 | 16,764 | -3.77% | 0.11% / 0.03% |
| doc_large | 89,651 B | 105,217 | 105,200 | 96,312 | 96,335 | -8.44% | 0.02% / 0.02% |

Decode, measured by pointing **one** binary at stores written by different arms so the
frame is the only variable. The null control is a **second, independently seeded
base-written store** of the same corpus — byte-identical frames in a different directory.
Each store runs in each of three positions, 9 inner reps:

| store | pos 0 | pos 1 | pos 2 |
|---|---:|---:|---:|
| L1 (base) | 7,141 | 7,103 | 7,105 |
| L1 (null control, 2nd seed) | 7,128 | 7,104 | 7,169 |
| L3 (candidate) | 5,474 | 5,465 | 5,401 |

Six identical-code observations span 7,103-7,169 (**0.93%**, the floor); the candidate's
median is 5,465. **Decode is -23.2%**, 25x the floor.

So the earlier characterisation was wrong twice: decode is not "unchanged", it is
substantially *faster*; and encode is not uniformly cheaper, as §1 shows.

### 7. Why the synthetic corpora mispredicted the real workload

This is the part worth carrying forward.

On six synthetic corpora, level 3 encode is **2.5-8.4% cheaper** than level 1, against a
null control of 0.02-0.22%. On the real commit path in `tracked_state_crud`, level 3 costs
**+231 µs** (rocksdb) and **+256 µs** (slatedb) per `update_all` operation. Opposite sign.

It is not compressor-context overhead: the oracle's `zstd_compress` calls
`zstd::bulk::compress`, which constructs a **fresh context per payload**, so context cost
is already inside the microbenchmark — and production is *more* favourable than that,
because `JsonBatchCompressor` reuses one context for a whole batch. The microbenchmark
gives level 3 the harder job and it still wins.

What is left is **payload shape**. The corpora in `expu_json_delta` are synthetic generated
documents; the JSON the engine actually writes on the commit path in this workload is
pnpm-lock-derived aggregate structure. Level 1's fast match finder and level 3's
greedy/lazy search rank differently on the two, and the synthetic ones do not stand in for
the real ones.

**The corpora, not the conclusion, are the thing to fix.** Any future compression-parameter
experiment on this store should be priced against payloads captured from the real commit
path, not against `doc_*`/`kv` generators.

### 8. Why not level 9 or 19 (for completeness)

Same null-control-and-rotation shape, `LIX_EXPU_ENCODE_LEVELS=1,3,9,19,19,9,3,1`:

| corpus | metric | L1 | L3 | L9 | L19 |
|---|---|---:|---:|---:|---:|
| doc_edits | ns/payload | 17,453 | 16,799 | 216,178 | 1,657,215 |
| doc_edits | vs L3 CPU | 1.04x | 1.00x | **12.9x** | **98.6x** |
| kv | ns/payload | 8,537 | 8,325 | 49,470 | 502,865 |
| kv | vs L3 CPU | 1.03x | 1.00x | **5.9x** | **60.4x** |

Level 9 buys a further 6.8-12.0% of bytes for 5.9-12.9x the encode CPU; level 19 for
60-99x. Both lose badly under the ranking. Given that even level 3 costs OLTP write time on
real payloads, nothing above 1 is currently defensible for this store.

## Correctness gate

Run on `ryzen-9950x-III` against `68ba7bc92`:

```
cargo check --workspace                                  ok
cargo clippy --workspace --all-targets --all-features -- -D warnings   ok
cargo test -p lix --all-features                         ok — 2922 passed, 0 failed, 13 ignored
cargo check -p lix_js_sdk --target wasm32-unknown-unknown ok
```

`--all-features` matters here because `packages/lix/src/storage_bench.rs` is gated behind
`storage-benches` and a plain `cargo test -p lix` never compiles it. The wasm target is
checked because `json_store/compression.rs` has `wasm32`-only branches that
`cargo check --workspace` cannot see.

## Recommendation

Do not merge the level change. Two things are worth salvaging as follow-ups:

1. **Keep the harness commits.** `LIX_EXPU_ENCODE_LEVELS` / `LIX_EXPU_ENCODE_DELTA` and the
   `today_bytes` fix are what turned "level 3 is faster" into a falsifiable claim. Without
   the null control this branch would have landed on the strength of a microbenchmark that
   was measuring the wrong payloads.
2. **Re-cut the corpora against real commit-path payloads** before anyone revisits
   compression parameters for the JSON store. The -23% decode result is real and large
   enough to be worth chasing on its own, but only if the encode side can be made to not
   cost OLTP write latency on the payloads the engine actually writes.


samuelstroschein:
Closing on the measurement. **I recommended landing this as "strictly dominant" and that was wrong** — it is not dominant on the current head, and the re-measurement is the deliverable.

## The decisive number

`update_all_rows` regresses **+2.98% (rocksdb) / +3.30% (slatedb)** at 9 reps, rotated order, with **completely non-overlapping raw ranges** (base 7.667–7.811 vs cand 7.901–8.032 ms). The null control — the candidate worktree rebuilt at base's own commit, byte-identical source — puts that same lane at **+0.21% / +0.55%**. Cleanest isolation: same worktree, same target dir, `update_all_rows` was 7.778/7.813 ms at level 1 and 7.994/8.011 ms at level 3.

Under `OLTP > version control > physical bytes`, buying 3–9% of bytes with a 3% OLTP write regression is the wrong direction, and nothing end-to-end clears 10% to pay for it.

## Why the earlier numbers were wrong — the reusable lesson

Both prior claims failed re-measurement, **in opposite directions**:

- **Encode was not cheaper.** On six synthetic corpora level 3 measured 2.5–8.4% cheaper (null spread 0.02–0.22%). On the **real commit path** it costs **+231/+256 µs per operation**. Not context overhead — the oracle already builds a fresh context per payload, which is harder than production's per-batch reuse. It is **payload shape**: the `doc_*`/`kv` generators do not stand in for the pnpm-lock-derived JSON the engine actually writes.
- **Decode was not unchanged** — it is **−23.2%** (5,465 vs 7,116 ns) against a 0.93% null floor. But it is invisible end-to-end (`read_all_rows_consumed` +0.28%/+0.37%) because only 22 JSON rows are involved.

**A compression level that is free on one payload population is not free on another.** That is why the level-1 call sites outside the JSON store (tracked-state chunks and parts, mutation journal, columnar row groups, plugin pages, wasm component — 256 KiB to 256 MiB payloads) were deliberately left untouched: there was no measurement for them, and this result vindicates that caution.

## Byte deltas, for the record

doc_small −2.79%, kv_rewrite −7.11%, kv −7.18%, doc_edits −7.31%, doc_header_only −7.31%, doc_scattered −7.31%, doc_large −8.78%. With the caveat the agent flagged unprompted: **doc_scattered is not independent** — its settled store is byte-identical to doc_edits, so that is six independent corpora, not seven.

Settled bytes moved in exactly one space (`json_store.json` 151,126 → 148,322, −1.86%); every other space, staged put/delete count, touched-space count, storage call and batch count is identical to the byte.

The branch is left pushed as reproducible evidence.

PR #1305, opened 2026-08-11
perf(branch): root-back explicit head moves instead of materializing the working set

samuelstroschein:
## Question this answers

*When you branch, how much storage do the branches share?* Nobody had measured it. This PR adds the
harness that measures it, and fixes the one place where the answer was "none".

## What changed physically

`stage_branch_head_control_publications` (`packages/lix/src/transaction/commit.rs`) had two arms for
an explicit branch-head publication:

* **new branch** (`existing.is_none()`) → `stage_root_backed_branch_publication`: publish one 16-byte
  reference to the immutable tracked root at the head. O(1).
* **existing branch** (head *moves* — which is what a fast-forward merge does) → load
  `HotTrackedSnapshot` for the head and call `stage_complete_current_state_with_working_diff`, which
  writes **one `HOT_ROW` per tracked entity in the whole workspace**. O(state).

Both arms are now the root-backed one. The existing-branch bookkeeping the second arm carried
(`current_state_revision` increment, preserved `working_diff_checkpoint_commit_id`/`created_at`,
untracked-generation rollover, insert absence guards) moved into
`stage_root_backed_branch_publication`, which already took `previous_control: Option<_>`.

**Removed**: the ~120-line materializing arm; the `tracked_snapshots` parameter of
`stage_branch_head_control_publications`; the `StagedHotHeads::tracked_snapshots` field. Net
−156 lines in `commit.rs`. No shims, no dual path, no feature flag. Public API unchanged.

**Deliberate behaviour difference**: a root-backed publication sets `schema_presence_bloom` to all
ones (conservative "may contain any schema") instead of carrying the previous bloom forward. That can
only cost an extra probe, never hide a row.

## The measurement

New bench `packages/engine-benchmarks/benches/branch_storage_sharing.rs` (`harness = false`). It
seeds N rows through a real `SessionContext` SQL commit path, settles storage (memtable flush +
sleep), snapshots **both** a recursive filesystem byte walk and per-space
`rows / key_bytes / value_bytes` from `lix::storage_bench::layout_accounting`, applies one branching
scenario, and re-snapshots. Delta = the price of the scenario, attributed per storage space.

## A/B: settled logical bytes, delta vs the seeded base

Machine `ryzen-9950x-I`, interleaved base/cand, medians. Logical byte counts are deterministic —
RocksDB and SlateDB agree to within 1 byte on the same scenario, and per-rep values are identical.

### rocksdb

| scenario | rows | base Δbytes | cand Δbytes | delta |
[...]

PR #1328, opened 2026-08-11
fix(working-diff): branch-local edits lost their before image, making revert delete the row

samuelstroschein:
## Data loss: reverting a branch-local edit deleted the row

On a branch created by `create_branch`, the **first** mutation of a row that already existed at the checkpoint was recorded as if that row had just been *created*. The consequences at the public SQL surface:

- **`lix_revert` deleted the user's checkpointed value instead of restoring it.** The diff said `added`, so revert correctly undid an "add" — by removing the row. The prior value is gone.
- **Branch-local deletes were invisible.** Deleting a checkpointed row on a branch produced *no* `lix_working_diff` row at all, narrow or broad.
- **Edited rows were missing from broad `lix_working_diff` queries entirely.** An unfiltered `SELECT … FROM lix_working_diff` returned only newly inserted keys; every edited pre-existing row was absent.
- Narrow (`entity_pk`-filtered) queries did return the row, but as `diff_type = 'added'` with `before_change_id = NULL`.

### Scope — read this part

- **The trigger is `create_branch`, not merge.** It also reproduces through `switch_branch` into such a branch (same root cause, not an independent one). `main` is unaffected because it is not root-backed.
- **The wrong baseline is sticky.** A second `UPDATE` of the same row does not repair it.
- **It affects every identity without a branch-local HOT row**, not just the literally-first edit. In the regression test, two separate pre-existing rows edited after an unrelated insert are *both* missing from the working diff.
- **PR #1305's root-backed head moves did not widen this.** An explicit head move on an *existing* branch stages no new working-diff epoch, so the stored epoch's generation goes stale, `commit.rs` yields `working_diff_epoch = None`, the branch control drops its `working_diff_checkpoint_commit_id`, and `working_diff_for_control` bails to the canonical diff. Post-merge working diffs are correct (unaccelerated). Our own recent work is not implicated.
- **Merge and merge preview were never affected** — they use the canonical commit diff, not the HOT working-diff accelerator. Two tests in this PR prove it and are expected to pass both before and after the fix.

## Root cause

One struct field conflated two different facts. `PackedHeadValue.checkpoint_commit_id: Option<CommitId>` was used for both current-state bases, and `CertifiedCurrentStatePredecessor::view()` mapped any `Some(..)` to `WorkingDiffBaseline::BeforeAbsent`:

```rust
// before
working_diff_baseline: value.checkpoint_commit_id.map_or(
    WorkingDiffBaseline::Disabled,
    |checkpoint_commit_id| WorkingDiffBaseline::BeforeAbsent { checkpoint_commit_id },
),
```

That is right for the **packed** current base — a collection published *inside* the working interval, therefore absent at the checkpoint. It is wrong for the **root** current base, which *is* the checkpoint state. `next_hot_working_diff_baseline` then returned `(BeforeAbsent, newly_dirty = false)`, which is both halves of the bug: the wrong classification, and no sparse dirty-key index entry — which is why broad queries omitted the row while coverage stayed self-consistent and nothing tripped the canonical fallback.

This is the third defect in this family: a type that can represent "unknown" or "not applicable" collapses onto the dangerous answer by default. The fix makes the unclassified state **unrepresentable as an answer** rather than defaulting.

## The fix

**Split the type.** `checkpoint_commit_id: Option<CommitId>` becomes:

```rust
enum PackedWorkingDiffBaseline {
    Disabled,
    AbsentAtCheckpoint { checkpoint_commit_id: CommitId },  // packed base
    CleanAtCheckpoint,                                       // root base
}
```

`view()` matches this exhaustively by name. There is no `_ =>` arm and no `map_or` anywhere on the path, so a new base kind is a compile error rather than a silent `BeforeAbsent`.

**Capture the before image by reference; resolve it lazily.** A baseline taken from a base that was read without payload projection records its payload slots as `WORKING_DIFF_SLOT_UNRESOLVED` (new kind `3`, zero hash, validated on decode). The write path stores only the reference it already had — change id plus owning commit — and does **no** new I/O.

**Make "unclassified" a distinct answer.** `payload_eq -> bool` becomes `payload_equality -> WorkingDiffPayloadEquality {Equal, Different, Unresolved}`. `Equal` still short-circuits on change id, so ordinary modifications never hydrate anything. `Unresolved` reaching classification is a hard error, not a guess.

**Hydrate on the read path.** `resolve_working_diff_before_payloads` runs before each classifier, groups unresolved rows by owning commit, and issues one `load_commit_delta_change_records` batch per commit. Identity keys are materialized only for rows that are actually unresolved, so a diff with no root-backed baselines allocates nothing extra and returns immediately.

> **Note for whoever touches this next.** My first implementation resolved via `load_change_records` against `CHANGE_SPACE` and failed with `loaded 0 of 1`. Change records are addressed **by owning commit**, not by raw change id. `WorkingDiffVersion` already carries `commit_id`, so the stored reference was complete — the lookup was wrong. That is why hydration batches per commit rather than per change.

## Layout invariants

1. **Single authority** — no new authority. One field replaced by one better-typed field; the root base remains the only source of its own baseline.
2. **Commit canonicality** — unchanged.
3. **Derived views are caches** — this restores the property. The HOT working-diff index is an accelerator over the canonical diff; it was answering *confidently and wrongly* where it should have answered correctly. It now either answers correctly or errors.
4. **Atomic publication** — unchanged; the baseline publishes in the same write set as the row.
5. **GC from refs** — unchanged. Before images are referenced through commits already reachable from the checkpoint.
6. **SQL reads canonical records** — strengthened: the accelerator now hydrates from the canonical change record when the reference alone is insufficient.

## Correctness

`cargo test -p lix --all-features` green (2073 + 847 + 12 across targets, 0 failed). `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean. `cargo check --workspace` clean.

Ten regression assertions, all **verified failing against the unfixed tree**, in both simulation variants (`_base`, `_tracked_state_rebuild`):

| test | pre-fix |
|---|---|
| `working_diff_first_branch_edit_of_existing_row_is_modified` | FAIL — `added`, before `NULL` |
| `revert_of_first_branch_edit_restores_the_checkpoint_value` | FAIL — **row deleted** |
| `working_diff_first_branch_delete_of_existing_row_is_removed` | FAIL — 0 rows, narrow **and** broad |
| `working_diff_branch_edit_is_modified_for_broad_and_repeated_edits` | FAIL — both edited rows missing |
| `working_diff_first_edit_after_switch_branch_of_existing_row_is_modified` | FAIL — `added` |
| `working_diff_first_edit_after_merge_of_existing_row_is_modified` | pass (guard: merge unaffected) |
| `merging_a_first_branch_edit_reports_modified_and_lands_the_value` | pass (guard: merge unaffected) |
| `checkpoint_after_first_branch_edit_keeps_the_edited_value` | pass (guard) |
| `working_diff_restoring_the_checkpoint_payload_on_a_branch_is_net_empty` | passed **vacuously** pre-fix (row was invisible); now a real assertion that exercises hydration |

That last one is the test that fails if anyone regresses hydration back to change-id-only equality.

## Cost — measured

The fix makes a branch-local first mutation stage a dirty-key index entry that `newly_dirty = false` wrongly skipped, and store `BeforePresent` (checkpoint id + 115-byte version) instead of the smaller `BeforeAbsent`. **That is the write that makes the row visible in the working diff at all**, and it is what a non-root-backed branch already pays.

`branch_storage_sharing`, `LIX_BRANCH_SHARING_ROWS=10000`, byte-exact and deterministic (no rep noise). Logical bytes per changed row, rocksdb — slatedb agrees to within 0.1%:

| scenario | changed rows | base | cand | delta |
|---|---:|---:|---:|---:|
| `branch_1row` | 1 | 5525.00 | 5760.00 | +235 (+4.3%) |
| `branch_1pct` | 100 | 417.32 | 583.58 | **+166.3 (+39.8%)** |
| `branch_10pct` | 1000 | 372.31 | 537.47 | **+165.2 (+44.4%)** |
| `merge_1pct` | 100 | 424.74 | 591.00 | **+166.3 (+39.1%)** |

**+166 logical bytes per branch-local changed row, flat across scale.** That decomposes cleanly: `WORKING_DIFF_VERSION_BYTES` = 115, plus ~50 bytes of dirty-key segment entry. Physical bytes move less: `branch_1pct` +5.8% rocksdb / +7.4% slatedb, `branch_10pct` +15.0% / +15.2%, `branch_1row` +0.3%.

Compounding note for the HOT-plane work: this moves the branch-local edit cost from ~390 B/row to **~556 B/row** against ~32 B in the packed plane. The fix is not the cause of that gap, but it widens it, and that argues for the HOT-plane layout work rather than against this fix.

Write latency (`scenario_ms`, single rep, noisy): `branch_1pct` +3.5%, `branch_10pct` −3.7%, `merge_1pct` −4.9%, `branch_1row` −17%. Straddles zero; no systematic write regression.

## Guards — honest status

`tracked_state_crud` (read + write ids), `untracked_state_crud` as the null control, and `tracked_working_diff`, interleaved with rotated arm order. At n=3 the run is **not resolvable**:

| | n | median | min | max |
|---|---:|---:|---:|---:|
| **null control** (`untracked_state_crud`, byte-identical code both arms) | 21 ids | −0.59% | −11.3% | **+25.5%** |
| `tracked_state_crud` | 84 ids | −0.85% | −21.0% | +21.5% |

The null control's own spread is ±25%, so every per-id delta in the guard set sits inside the noise floor. The two medians (−0.85% vs −0.59%) are indistinguishable. The defensible statement is: **no systematic shift, and this harness at n=3 cannot resolve anything smaller than roughly ±25%.**

This is expected rather than lucky. The change is provably inert off the branch path: with no root current base, `CleanAtCheckpoint` is never constructed, `payload_is_unresolved()` is always false, and the resolver returns at `pending.is_empty()` before touching storage. The benches above all run on `main`.

A longer 5-rep run was still in flight when this was opened; the deterministic byte accounting above is the load-bearing cost number and is not subject to rep noise. I will post the completed timing table as a comment.

### DIFF
--- a/.changenotes/fix-branch-working-diff-baseline.md
+++ b/.changenotes/fix-branch-working-diff-baseline.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+Fixed branch-local edits being reported as additions in the working diff, which made reverting one delete the row instead of restoring its previous value.
+
+On a newly created branch, editing a row that already existed at the checkpoint was recorded as if the row had just been created. `lix_working_diff` reported it as `added` with no before value, deleting such a row did not appear in the working diff at all, and reverting the edit removed the row rather than restoring the checkpointed value. Branch-local edits now carry their correct before image, so working diffs classify them as modifications and reverts restore the previous value. Merges, merge previews and the main branch were never affected.
--- a/packages/lix/src/live_state/tracked_head.rs
+++ b/packages/lix/src/live_state/tracked_head.rs
@@ -182,6 +182,15 @@ struct WorkingDiffSlotFingerprint {
 const WORKING_DIFF_SLOT_NONE: u8 = 0;
 const WORKING_DIFF_SLOT_REF: u8 = 1;
 const WORKING_DIFF_SLOT_INLINE: u8 = 2;
+/// The before image is identified by its change id, but its payload slot was
+/// not materialized where the baseline was captured.
+///
+/// Root current bases are read under `ChangeRecordProjection::identity_only()`
+/// so the write path pays no payload I/O to capture a baseline. The reader
+/// hydrates the referenced change record only for the one question the change
+/// id cannot answer on its own: whether two distinct changes carry the same
+/// payload.
+const WORKING_DIFF_SLOT_UNRESOLVED: u8 = 3;
 const WORKING_DIFF_VERSION_BYTES: usize =
     16 + 16 + 1 + 8 + 8 + 1 + JSON_REF_BYTES + 1 + JSON_REF_BYTES;
 const WORKING_DIFF_CHECKPOINT_BYTES: usize = 16;
@@ -1300,16 +1309,21 @@ fn decode_working_diff_slot(
         .ok_or_else(|| working_diff_error(&format!("{field} kind is missing")))?;
     if !matches!(
         kind,
-        WORKING_DIFF_SLOT_NONE | WORKING_DIFF_SLOT_REF | WORKING_DIFF_SLOT_INLINE
+        WORKING_DIFF_SLOT_NONE
+            | WORKING_DIFF_SLOT_REF
+            | WORKING_DIFF_SLOT_INLINE
+            | WORKING_DIFF_SLOT_UNRESOLVED
     ) {
         return Err(working_diff_error(&format!("{field} slot kind is invalid")));
     }
     let hash: [u8; JSON_REF_BYTES] = take_working_diff_bytes(bytes, offset, JSON_REF_BYTES)?
         .try_into()
         .map_err(|_| working_diff_error(&format!("{field} hash is invalid")))?;
-    if kind == WORKING_DIFF_SLOT_NONE && hash != [0; JSON_REF_BYTES] {
+    if matches!(kind, WORKING_DIFF_SLOT_NONE | WORKING_DIFF_SLOT_UNRESOLVED)
+        && hash != [0; JSON_REF_BYTES]
+    {
         return Err(working_diff_error(&format!(
-            "{field} none slot must have a zero hash"
+            "{field} slot kind must have a zero hash"
         )));
     }
     Ok(WorkingDiffSlotFingerprint { kind, hash })
@@ -1409,6 +1423,11 @@ struct HeadValueView<'a> {
     metadata: HeadSlotView<'a>,
     columnar_base_coordinate: Option<ColumnarBaseCoordinate>,
     working_diff_baseline: WorkingDiffBaseline,
+    /// False when the JSON slots above are placeholders because this view was
+    /// synthesized from a current-state base that was read without payload
+    /// projection. Baselines captured from such a view record their payload
+    /// slots as unresolved and are hydrated by the reader on demand.
+    payload_slots_materialized: bool,
 }
 
 impl CertifiedCurrentStatePredecessor {
@@ -1426,15 +1445,12 @@ impl CertifiedCurrentStatePredecessor {
                 deleted: value.deleted,
                 created_at: value.created_at,
                 updated_at: value.updated_at,
-                // Neither base materializes its JSON slots here. A packed base
-                // has no active-epoch before image at all, and a root base
-                // relies on `WorkingDiffVersion::payload_eq` short-circuiting
-                // on change id: a genuine mutation always mints a new change
-                // record. The residual cost is that restoring the checkpoint
-                // payload on a root-backed generation reports `modified`
-                // rather than a net no-op.
+                // Current-state bases are read without payload projection, so
+                // these slots are placeholders. The change id above is the
+                // reference the reader hydrates from when it needs the payload.
                 snapshot: HeadSlotView::None,
                 metadata: HeadSlotView::None,
+                payload_slots_materialized: false,
                 columnar_base_coordinate: value.columnar_base_coordinate,
                 working_diff_baseline: match value.working_diff_baseline {
                     PackedWorkingDiffBaseline::Disabled => WorkingDiffBaseline::Disabled,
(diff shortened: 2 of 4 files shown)


samuelstroschein:
## Guard run complete — no lane moves beyond the null control

Machine `hetzner-cpx62-III`. Arms: `~/claude3/base` (pre-fix) vs `~/claude3/cand` (this branch), interleaved with **rotated arm order** (odd reps base→cand, even reps cand→base). `untracked_state_crud` is the null control: byte-identical code in both arms, so its spread is this harness's resolution limit.

| lane | ids | reps/arm | median | min | max |
| **null control** `untracked_state_crud` | 21 | 5 | **−0.35%** | −15.2% | +12.6% |
| `tracked_state_crud` (read + write) | 84 | 5 | −0.70% | −21.0% | +13.9% |
| `diff_commands` | 4 | 3 | +0.21% | −1.95% | +1.10% |
| `tracked_working_diff` | 1 | 5 | +2.39% | — | — |

**Read:** the null control alone spans −15% to +13%, so every per-id delta in `tracked_state_crud` sits inside the noise floor, and its median (−0.70%) is indistinguishable from the control's (−0.35%).

`diff_commands` is the one instrument here with real resolution — criterion collects 100 internal samples per id — and it is **flat within ±2%** on every id, including the two that matter most for this change:

```
 -1.95%  base= 3.387ms  cand= 3.321ms  diff_commands/working_diff_1000
 -0.21%  base= 9.303ms  cand= 9.284ms  diff_commands/revert_100_of_1000
 +0.63%  base=18.934ms  cand=19.053ms  diff_commands/partial_checkpoint_500_of_1000
 +1.10%  base= 9.178ms  cand= 9.280ms  diff_commands/apply_100_of_1000
```

`tracked_working_diff`'s +2.39% is not a signal: the raw per-rep p50s are `base [0.683, 0.322, 0.502, 0.375, 0.528]` and `cand [0.514, 0.258, 0.511, 0.547, 0.617]` ms. The within-arm range is roughly twice the between-arm delta on a sub-millisecond measurement.

This matches the mechanism: the change is inert off the branch path. With no root current base, `CleanAtCheckpoint` is never constructed, `payload_is_unresolved()` is always false, and `resolve_working_diff_before_payloads` returns at `pending.is_empty()` before touching storage. All four benches above run on `main`.

The load-bearing cost number remains the deterministic byte accounting in the PR description (**+166 logical bytes per branch-local changed row**, byte-exact, no rep noise).

PR #1350, opened 2026-08-12
refactor(layering): delete the nine tracked_state -> live_state back-edges and enforce the order

samuelstroschein:
Follow-up to the `expmod` module-boundary audit. The audit's headline was **do not merge
`live_state` and `tracked_state`** — the merge would destroy the only structural enforcement of
layout invariant #3 — and its replacement recommendation was to delete the nine `tracked_state` →
`live_state` back-edges instead. This PR does that, and then declares and enforces the layering
those deletions make possible.

**This is a pure relocation. No logic changed, no signature changed beyond what the move forces, no
storage space moved owner, no write path was altered.** `entity_columnar.rs` moves with a
zero-line diff (git records it as a rename).

### 1. `live_state/entity_columnar.rs` → top-level `entity_columnar` (7 of 9 back-edges)

Verified against the tree: the file's 117 production lines import **only** `crate::changelog::CommitId`
and `crate::columnar_row_group::{…}`. It references nothing in `live_state`, and its own doc comment
says it is "layered over generic immutable columnar row groups". It was misfiled.

It is also *shared*: `sql2`, `transaction`, `gc`, `session`, `tracked_state` and `live_state` all use
it. That makes it common vocabulary below both state planes, not hot-plane machinery.

It did **not** go into `columnar_row_group` as the audit suggested, because it depends on
`changelog::CommitId` (it addresses row groups by owning commit) and `columnar_row_group` today does
not — putting it there would push a canonical-record dependency into a module the layering describes
as "pure encodings, no repository semantics". A top-level module above `changelog` and below both
planes is the honest placement, and `entity_pk` is the existing precedent for a small vocabulary
module at that level.

`live_state/mod.rs` keeps a documented `pub(crate) use crate::entity_columnar::{…}` facade so the
move did not have to edit every call site. That was deliberate: two of the consumers are
`transaction/commit.rs` and `sql2/providers/entity.rs`, which other agents are actively editing.
Those consumers all sit *above* `live_state` already, so the facade creates no new edge and distorts
no layer; it is removable churn, not debt with teeth.

### 2. `tracked_state::load_certified_rows_at_commit` → `live_state` (2 of 9)

Verified: the method's entire body is a call to `live_state::scan_certified_history_rows`, it returns
a `live_state` type, it touches nothing on `self` except `self.store`, and it has exactly **one**
production caller (`session/merge/branch.rs:963`).

It is now a free function in `live_state/tracked_head/hot.rs`, next to the scan it wraps, taking the
store directly. `TrackedStateStoreReader` gains a `store()` accessor so the single caller can run the
other plane's read against the same snapshot. Body copied verbatim; only the fully-qualified
`crate::tracked_state::` prefixes were dropped, since those names are already in scope at the new site.

### Result

`tracked_state` → `live_state` production references: **9 → 0.** The three remaining textual
references are all inside the `#[cfg(test)]` span beginning at `tracked_state/context.rs:4856`
(test doubles legitimately reach across planes; out of scope per the audit).

## Second deliverable: `module_layers.rs`

An in-crate `#[cfg(test)]` guard, modelled on `storage_spaces.rs::every_declared_space_is_registered`
— which already walks `CARGO_MANIFEST_DIR/src` at test time.

**In-crate rather than CI, and the reason is decisive: this fleet opens draft PRs, and `ci.yml` gates
on `draft == false`. A CI-only check would never run.** This one rides
`cargo test --workspace --all-features`, which every agent already runs.

- `MODULE_LAYERS: &[&[&str]]` is the declared order — the artifact itself, reviewable in a diff.
- The test attributes every `crate::<module>` path in **production** source (`#[cfg(test)]` items,
  comments and string/raw-string/char literals are blanked first) to the top-level module of its
  file, and fails on any reference to an equal or higher layer.
- `UNLAYERED_MODULES` names every top-level module *not* yet ordered, each with its reason, and a
  second test asserts the two lists together cover exactly what is on disk — so a new module cannot
  silently escape the guard, and the exception list is a visible to-do rather than an amnesty.
- `ALLOWED_UPWARD_EDGES` exists as the documented escape hatch and is **empty**: nothing in the
  declared order needs an allowance.

Declared order (bottom first), all 13 entries verified acyclic against the current tree:

| Layer | Modules |
|---|---|
| 0 | `compression`, `storage_codec` |
| 1 | `storage_adapter` |
| 2 | `binary_cas`, `columnar_row_group`, `common`, `json_store` |
| 3 | `entity_pk` |
| 4 | `changelog` |
| 5 | `entity_columnar` |
| 6 | `tracked_state` |
| 7 | `commit_graph` |
| 8 | `live_state` |

Two audit self-corrections are honoured: `commit_graph` sits **above** `tracked_state` (history
resolution genuinely needs the commit-delta plane — one edge, honestly placed, rather than an
exception), and `gc`/`session`'s mutual reference is the correct reclamation design, so both are
simply left unlayered rather than recorded as defects.

Nothing aspirational is encoded. `transaction`, `sql2`, `session`, `filesystem`, `plugin`, `branch`
and the rest stay in `UNLAYERED_MODULES` with the cycle named, because a guard that starts red is a
guard that gets `#[ignore]`d. The dominant blocker is recorded there too: 19 of the 20 inbound
back-references to `transaction` target `transaction::types`, so extracting it downward retires six
of the eleven cycles at once.

Two departures from the audit's layer table, both forced by measurement rather than taste:
`storage_adapter` sits **below** `common` and `columnar_row_group` (they reference it, not the
reverse), and `storage` is not a layer at all — `storage/mod.rs` re-exports `crate::handle` and
`storage/in_memory.rs` uses `crate::filesystem`, so it is a public API facade.

### The guard is not vacuous

Same analysis, same layer table, run against both trees:

```
BASE claude-4-optimizations: 1 violating edge
    tracked_state(L6) -> live_state(L8)  x9
CANDIDATE:                  0 violating edges
```

Nine — exactly the audit's count. The declared order does not merely describe the tree; it fails on
the parent commit and passes here. A third test asserts the scanner still finds >500 cross-module
references and still sees `live_state` → `tracked_state`, so the guard cannot pass by going blind.

## Layout invariant checklist

1. **Single authority** — no new authority. Nothing was duplicated; two items moved and one accessor
   was added. Space ownership is untouched: every hot-plane space is still written only from
   `live_state/tracked_head/hot.rs`, every canonical-state space only from `tracked_state/`.
2. **Commit canonicality** — unaffected; no commit or ref handling touched.
3. **Derived views are caches** — **strengthened, and this is the point of the PR.** The property
   "the cache may read the canonical store; the canonical store may not read the cache" was
   previously conventional and violated nine times. It is now structural: `tracked_state` cannot
   name `crate::live_state` without failing `no_module_references_a_higher_layer`.
4. **Atomic publication** — unaffected; no staging or publication code changed.
5. **GC from refs** — unaffected. (The audit separately refutes the "gc writes foreign spaces" claim
   on current code; P6 should be struck from the queue.)
6. **SQL reads canonical records** — unaffected; no provider logic changed.

## Gate

Full CI scope from the runbook plus features-off, on `hetzner-cpx62-IV` (verified idle, `pgrep -a
cargo` empty, reflog clean before starting). Actual output pasted in the thread.

**No timing run.** A pure relocation with a zero-line diff on the moved file cannot change a
benchmark; had one moved, the correct response would be to stop, not to report the number.

## Follow-ups this unblocks

- Drop the `live_state` re-export facade once `transaction/commit.rs` and `sql2/providers/entity.rs`
  are free.
- Extract `transaction/types.rs` downward — 6 of 11 cycles, the highest structural payoff per line
  in the audit — then move `transaction`, `sql2`, `branch`, `plugin`, `checkpoint`,
  `collection_generation` and `undo_redo` out of `UNLAYERED_MODULES`.
- Extract the duplicated dictionary/batch machinery into a shared L0 module (~555 lines, of which
  147 are provably byte-identical). Note the live side was tuned to `FastHashBuilder` and the
  tracked-side twin still uses `RandomState::with_seeds(0,0,0,0)` — one copy got the optimization and
  the other did not, which is the recurring cost of the duplication.


samuelstroschein:
## Gate output — hetzner-cpx62-IV, post-merge of `claude-4-optimizations`

HEAD: `e440ed687 Merge remote-tracking branch 'origin/claude-4-optimizations' into refactor/module-layering-backedges`

```
cargo check  --workspace --all-targets --all-features                                      EXIT=0
cargo clippy --workspace --all-targets --all-features -- -D warnings                       EXIT=0   (no warnings emitted)
cargo check  -p lix --no-default-features                                                  EXIT=0
cargo check  -p lix_js_sdk --target wasm32-unknown-unknown                                 EXIT=0
cargo test   --profile test --workspace --exclude lix_benchmarks --all-features
             --no-fail-fast                                                                EXIT=0
cargo test   --profile test -p lix_benchmarks
             --features storage-benches,slatedb,root-replay-trace --no-fail-fast           EXIT=0
```

Every `test result` line across both test runs reports `0 failed`; no `panicked`, `FAILED` or
`error` line appears anywhere in the log. Machine was verified idle before starting
(`pgrep -a cargo` empty) and `~/claude3/cand` reflog showed only this branch.

The guard in the `lix` lib target (2,098 tests):

```
running 4 tests
test module_layers::no_module_is_declared_twice ... ok
test module_layers::every_top_level_module_is_accounted_for ... ok
test module_layers::no_module_references_a_higher_layer ... ok
test module_layers::the_scanner_finds_the_references_it_is_supposed_to ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 2094 filtered out; finished in 1.35s
```

### Negative control: the guard was run red on purpose

Reintroducing the back-edge on the machine (`crate::entity_columnar::entity_row_group_set_id` →
`crate::live_state::…` in `tracked_state/storage.rs`, then reverted) makes the guard fail with the
exact call sites:

```
thread module_layers::no_module_references_a_higher_layer panicked at
packages/lix/src/module_layers.rs:499:5:
module layering violations (5 distinct):
  tracked_state/storage.rs:1065  — `tracked_state` (layer 6) references `live_state` (layer 8)
  tracked_state/storage.rs:3262  — `tracked_state` (layer 6) references `live_state` (layer 8)
  tracked_state/storage.rs:6848  — `tracked_state` (layer 6) references `live_state` (layer 8)
  tracked_state/storage.rs:10482 — `tracked_state` (layer 6) references `live_state` (layer 8)
  tracked_state/storage.rs:13226 — `tracked_state` (layer 6) references `live_state` (layer 8)
```

Five, not seven, because the other two named `entity_identity_column_index`, which the
`live_state` facade no longer re-exports — so those two back-edges now fail to *compile* rather
than fail the test. Both halves of the boundary are enforced.

**No timing run**, per the reasoning in the description: `entity_columnar.rs` moves with a zero-line
diff and `load_certified_rows_at_commit` keeps a verbatim body, so there is no mechanism by which a
benchmark could move. If one did, that would mean the move was not pure and the right response
would be to stop, not to publish a number.

PR #1353, opened 2026-08-12
feat(plugin): run plugins on untracked files

samuelstroschein:
PR **C of the untracked-lane chain**: A (propagation, #1332) → B (change ids, #1346) → **C (unskip)** → D (enforcement).

Owner ruling being implemented: **plugins should behave identically, irrespective of tracked or untracked.** Before this PR, plugin reconciliation was skipped for untracked writes, so an untracked JSON file was a descriptor plus an opaque content blob with **no entity rows at all** — its contents were unqueryable. The motivating case is cache files that happen to be JSON.

## What changed

**1. Reconciliation no longer skips untracked file writes.** Every file-scoped read on the plugin path now follows the file's own lane instead of a hardcoded tracked constant: candidate selection, owner lookup, descriptor tombstones, plugin selection, state-group scans, the cold-open path, entity hydration, create authorities, and semantic write-back.

**2. Untracked complete parses are materialized as ordinary changes, not certified packets.** This is the substantive design decision.

A component plugin's created entities leave the guest in a host-synthesized **certified entity packet**. `transaction/commit.rs` materializes those rows by expanding them **from the file's published commit root**:

```rust
let root = tracked_roots.iter().find(|r| r.publish_head && r.branch_id == file.branch_id)
    .ok_or_else(|| LixError::new(CODE_INTERNAL_ERROR, "certified entity batch has no matching published commit"))?;
```

An untracked file publishes no commit, so there is no root to expand from. `WasmOpenFileInput` now carries `certified_packets_available`; when false the same packet payload takes the pre-existing ordinary decode branch. **Same guest entities, row-shaped instead of packed.** No new artifact, no new storage space, no head-encoding change, no new code path — it selects an existing branch of existing code.

Two alternatives were priced and rejected. An *untracked certified root* (a packed root not keyed to a commit) invents an artifact whose failure mode is silent cross-lane corruption. An *untracked durable checkpoint* does not address the failing path at all: a create has no prior checkpoint by construction.

**3. Two lane-per-batch hazards fixed rather than deferred.** `LiveStateExactBatchRequest` carries one lane per batch. Both `preflight_creates` and the plugin **owner lookup** are now lane-partitioned. The owner lookup was not in the inventory and had the same defect: read under a constant lane it would report an untracked file as unowned and re-own it on every write.

**4. The read-side mirror of the #1332 blocker.** `validate_live_state_identity` rejected *any* untracked reserved plugin row, so an untracked file's owner row could be written but never read back. Caught by the fresh-reparse test, not by inspection.

## Evidence

All four tests **fail on unmodified `claude-4-optimizations`** (run there with only the test file applied) and pass here.

| test | on base | here |
|---|---|---|
| `untracked_json_file_produces_the_same_plugin_entity_rows_as_a_tracked_one` | 0 entity rows | ok |
| `untracked_plugin_rows_enforce_foreign_keys_like_tracked_ones` | untracked graph `[]` vs tracked 8 nodes | ok |
| `a_fresh_reparse_after_ownership_loss_replaces_untracked_rows_like_tracked_ones` | `[]` vs `["alpha"]` | ok |
| `untracked_file_matching_no_plugin_is_unaffected_by_reconciliation` | ok (regression guard) | ok |

The acceptance test keeps the tracked arm **beside** the untracked one so "identical irrespective of lane" is asserted rather than asserted-about, and asserts the #1346 shape: a real `change_id` with a NULL `commit_id`.

**The two equivalence questions were settled by execution, not argument**, because the decode path uses different machinery from the certified path:

- **Foreign keys.** A certified packet checks FKs *within the batch*; the decode path resolves them against live state. `markdown_node` (self-referential FK `/parent_id` → `/id`) produces an identical node graph on both lanes, every FK resolves in both, and deleting a referenced parent is rejected with the same error code on both.
- **Collection replacement.** The `complete_file_state` replacement marker is not produced on the decode path. Tested where that would matter — a file reparsed fresh after ownership loss, with prior rows present. Tombstoning and the decode path compose: the superseded member is gone on both lanes.

## Known characteristics and scope boundaries

- **Untracked files are permanently cold-start.** They get no durable plugin checkpoint, so a large untracked file is re-parsed on every open. This is the honest price of the lane boundary and is stated in the changenote.
- **Follow-up worth taking (not a blocker):** giving untracked files a checkpoint is a clean, independent perf win. `PLUGIN_CHECKPOINT_SPACE` is a dedicated side space with no lane, no commit and no changelog; it is content-addressed so a stale entry self-invalidates into a cold rebuild, and `stage_delete_branch_plugin_checkpoints` already reclaims it by branch prefix. It satisfies invariant 3 by construction. It is *not* the cross-lane state the original ruling assumed.
- **Two paths deliberately left tracked-pinned, both commented in code:** `preflight_owned_generation_upgrades` (validation-only, writes no rows — an untracked owned file is validated late rather than at upgrade time), and the plugin render context in `sql2/providers/file.rs` (resolves owners through the tracked-only `scan_tracked_batch`). Untracked file bytes round-trip without the latter, which the acceptance test asserts by editing an entity row and reading the file back.
- **Layering (#1350):** `transaction`, `plugin`, `wasm` and `default_wasm_runtime` are all in `UNLAYERED_MODULES`, so no constrained edge is added and `ALLOWED_UPWARD_EDGES` stays empty.
- Not asserted as an engine invariant, changenote guidance only: untracked change ids identify a row within its writing session and should not be compared across sessions.

## Gate

Full CI scope plus features-off, on `ryzen-9950x-V`. Actual output in the PR thread.


samuelstroschein:
## Gate — actual output, `ryzen-9950x-V`, on `e65d9e549` (merged with `claude-4-optimizations` @ `f810d00a6`)

```
cargo check  --workspace --all-targets --all-features   Finished in 30.61s   (clean)
cargo clippy --workspace --all-targets --all-features -- -D warnings
                                                        Finished in 53.85s   (clean)
cargo check  -p lix --no-default-features               Finished in 26.28s   (clean)
```

`cargo test --profile test --workspace --exclude lix_benchmarks --all-features --no-fail-fast` — **zero failures across every target**:

```
test result: ok. 2093 passed; 0 failed; 8 ignored     (lix lib)
test result: ok.  862 passed; 0 failed; 2 ignored     (lix integration)
test result: ok.   95 passed; 0 failed; 4 ignored     (lix-server-protocol)
test result: ok.   79 passed; 0 failed; 18 ignored
test result: ok.   67 passed; 0 failed                (x2)
test result: ok.   39 passed; 0 failed
... 54 targets total, all ok
```

`cargo test --profile test -p lix_benchmarks --features storage-benches,slatedb,root-replay-trace --no-fail-fast`:

```
test result: ok. 4 / 2 / 2 / 2 / 2 / 0 / 0 / 2 / 2 / 10 passed; 0 failed
```

Note `same_base_remote_plugin_writes_resolve_and_converge` (the known ~12-28% pre-existing flake) did **not** fire this run — server-protocol is 95/95.

## Base proof

Ran on a throwaway branch = unmodified `claude-4-optimizations` (`f810d00a6`) with **only** the test file applied:

```
untracked_json_file_produces_the_same_plugin_entity_rows_as_a_tracked_one
  expected exactly one plugin entity row for file '...8a2', got 0
untracked_plugin_rows_enforce_foreign_keys_like_tracked_ones
  left: []   right: [("document", Null), ("heading", ...), ... 8 nodes]
a_fresh_reparse_after_ownership_loss_replaces_untracked_rows_like_tracked_ones
  left: []   right: ["alpha"]
untracked_file_matching_no_plugin_is_unaffected_by_reconciliation ... ok

test result: FAILED. 1 passed; 3 failed
```

All three capability tests fail on base for the right reason — the rows do not exist at all — while the regression guard passes on both arms.

PR #1356, opened 2026-08-12
feat(state): a row and the file that owns it must share one lane

samuelstroschein:
**PR D, the last of the untracked-lane chain**: A (#1332) → B (#1346) → C (#1353) → **D**.

The ordering rule that produced the split — *enforcement without propagation must never reach the branch* — is satisfied: A, B and C are all in. D makes the invariant **structural rather than conventional**.

## The invariant

**A row and the file that owns it are in the same lane.** Before this PR an untracked row could name a tracked file as its owner, and that combination was quietly lossy: deleting the file, or undoing the commit that created it, took the untracked rows with it.

## What changed

**1. Enforcement — one function.** `Domain::file_owner_domains()` returns `vec![self.clone()]` instead of `reachable_target_domains()`.

Scoped to that function **only**, and this is load-bearing: `reachable_target_domains()` is shared with `fk_target_domains()` and `directory_parent_domains()`, where an untracked row referencing a tracked schema, account or parent directory is legitimate and must keep working. File ownership is the single relationship that must not cross the lane boundary. The function now carries a comment saying so, because the obvious "simplification" is to push the change down into `reachable_target_domains()` and that would break FK and directory scoping.

**2. A real two-direction lane-mismatch error.** Both directions previously reported `CODE_FILE_NOT_FOUND` — "references missing file_id" — for a file that `SELECT * FROM lix_file` plainly returns, which sends people hunting the wrong bug. When the own-lane lookup fails, validation now probes the other lane (pending and committed) and, if the file is there, returns `CODE_CONSTRAINT_VIOLATION` naming the row's lane and the file's lane:

> file ownership validation failed for schema 'x': untracked entity '["post-1"]' references file_id '…', which exists but is tracked, on branch '…'

A genuinely absent file still reports `CODE_FILE_NOT_FOUND`.

**3. `reject_untracked_descriptor_cascade` deleted** (`session/undo_redo.rs`).

**Unreachability argument, not just a green suite.** The guard existed to stop undo/redo removing a file descriptor whose file owned untracked rows. It scanned for untracked rows owned by files whose descriptors the delta removes. After (1), such a row cannot be written: the only way to create one is an INSERT naming a tracked `lixcol_file_id` with `lixcol_untracked = true`, and validation now rejects exactly that, in both the pending and committed descriptor cases (four tests). The guard's predicate is therefore unsatisfiable by construction rather than merely unobserved. Its two `Transaction` helpers, `has_untracked_file_scoped_rows` and `has_untracked_rows`, had no other callers and are removed with it; `apply_state_diff` loses the `target` parameter it only needed to build the guard's error message.

The removal is also a behaviour *restoration*: undo/redo previously refused to run at all whenever the mixed state existed. The converted test asserts undo now completes normally.

**4. The cross-lane cascade branch in `hot.rs` is KEPT, as a defensive skip.** Both cascade sites now skip on `delta.untracked != existing.untracked` rather than only the one direction.

**Do not delete the `existing.untracked` branch inside them as dead code.** It is the *only* mechanism that removes an untracked file's own rows when that file is deleted. Both lanes share one path keyed on the row's own flag, so what this invariant kills is the cross-lane *pairing* — which is the skip condition, not the branch. Both sites carry that note in code.

## Evidence

All five tests **fail on unmodified `claude-4-optimizations`** (`860123b2d`), run there with only the test bodies applied — production untouched, verified by grep before the run (guard still present, new error absent, `file_owner_domains` unchanged):

```
failures:
    session::undo_redo::tests::untracked_row_cannot_be_owned_by_a_tracked_file
    transaction::validation::tests::validation_rejects_tracked_file_owner_reference_committed_only_as_untracked
    transaction::validation::tests::validation_rejects_tracked_file_owner_reference_pending_only_as_untracked
    transaction::validation::tests::validation_rejects_untracked_file_owner_reference_committed_as_tracked
    transaction::validation::tests::validation_rejects_untracked_file_owner_reference_pending_as_tracked
test result: FAILED. 2088 passed; 5 failed
```

Two of those are the inversions called for in the plan. The other two are their **pending-descriptor siblings**, which asserted the same removed behaviour for a descriptor staged in the same transaction; the plan named only the committed pair, and the gate found the rest.

`undo_file_creation_rejects_deleting_untracked_file_state` became `untracked_row_cannot_be_owned_by_a_tracked_file`: it asserts the **INSERT** is rejected and names the file's lane, then asserts undo runs normally afterwards.

## Explicitly NOT closed by this PR

- **Recursive directory delete is lane-scoped** (`filesystem/planner.rs:1658`) and silently orphans untracked files under a deleted tracked directory — the inverse asymmetry to the file cascade. Pre-existing and independent; **D does not close it**, and nobody should read this PR as having settled directories.
- **`preflight_owned_generation_upgrades`** remains tracked-pinned, with its comment from #1353.

## Gate

Full CI scope plus `cargo check -p lix --no-default-features`, on `ryzen-9950x-V`. Output in the thread.


samuelstroschein:
## Gate — actual output, `ryzen-9950x-V`, on `c5a3bbcbb` (base `claude-4-optimizations` @ `860123b2d`)

```
cargo check  --workspace --all-targets --all-features    clean (no errors, no warnings)
cargo clippy --workspace --all-targets --all-features -- -D warnings
                                                         clean
cargo check  -p lix --no-default-features                Finished in 3.24s
```

`cargo test --profile test --workspace --exclude lix_benchmarks --all-features --no-fail-fast` — **54 targets, all ok, zero failures**:

```
     16 test result: ok. 0 passed; 0 failed
      4 test result: ok. 1 passed; 0 failed
      2 test result: ok. 4 passed; 0 failed
      2 test result: ok. 3 passed; 0 failed
      1 test result: ok. 862 passed; 0 failed; 2 ignored    (lix integration)
      1 test result: ok.  95 passed; 0 failed; 4 ignored    (lix-server-protocol)
      1 test result: ok.  79 passed; 0 failed; 18 ignored
      1 test result: ok.  67 passed; 0 failed
      ...
grep -c '^test result: ok' = 54     grep -c 'FAILED' = 0
```

`cargo test -p lix --lib --all-features`: `2093 passed; 0 failed; 8 ignored`.

`cargo test --profile test -p lix_benchmarks --features storage-benches,slatedb,root-replay-trace --no-fail-fast`: no failures.

`same_base_remote_plugin_writes_resolve_and_converge` (known ~12-28% pre-existing flake) did not fire — server-protocol is 95/95.

## Base proof

Unmodified `claude-4-optimizations` (`860123b2d`) with **only the test bodies** applied. Verified before the run that production was untouched:

```
reject_untracked_descriptor_cascade in undo_redo.rs : 3   (guard still present)
lane_mismatched_file_owner_error in validation.rs   : 0   (new error absent)
reachable_target_domains in domain.rs               : 5   (seam unchanged)
```

```
failures:
    session::undo_redo::tests::untracked_row_cannot_be_owned_by_a_tracked_file
    transaction::validation::tests::validation_rejects_tracked_file_owner_reference_committed_only_as_untracked
    transaction::validation::tests::validation_rejects_tracked_file_owner_reference_pending_only_as_untracked
    transaction::validation::tests::validation_rejects_untracked_file_owner_reference_committed_as_tracked
    transaction::validation::tests::validation_rejects_untracked_file_owner_reference_pending_as_tracked

test result: FAILED. 2088 passed; 5 failed; 8 ignored
```

Exactly the five, nothing else.

samuelstroschein:
## Re-gated after merging `claude-4-optimizations` @ `c055e1c9e`

That merge included `refactor/promote-transaction-types`, which moves `transaction::types` to a top-level `transaction_types` module — it touches the same files this PR changes, so the gate was re-run rather than assumed. Branch head is now `9406cb867`.

```
cargo clippy --workspace --all-targets --all-features -- -D warnings   clean
cargo check  -p lix --no-default-features                              Finished in 31.74s
cargo test   --profile test --workspace --exclude lix_benchmarks --all-features --no-fail-fast
    grep -c '^test result: ok'  = 54
    FAILED / failures: / error[] = none
```

Merge was clean and no conflict resolution was needed in the files this PR touches.

PR #1399, opened 2026-08-12
perf(sql): cut three per-statement incidental costs on the point-read path

samuelstroschein:
# perf(sql): cut three per-statement incidental costs on the point-read path

E38's differential budget itemized a 29.19 µs point read on this host (ryzen-9950x-I, RocksDB,
10k rows) of which ~18 µs is per-statement machinery above the store. This PR removes three of the
ranked incidental items, cheapest-per-risk first. Nothing about the public API, SQL semantics, or
result ordering changes.

## What changed

**1. Session recycle is O(1) instead of a catalog enumeration (E38 item: 2.6 µs).**
`ReadSqlSession::drop` recycled the pooled DataFusion session by calling
`MemorySchemaProvider::table_names()` — allocating one owned `String` per registered table by
walking every dashmap shard — then deregistering names one at a time through three-part
`TableReference`s. The enumeration is *not* pure waste: it exists so a later statement cannot
resolve stale snapshot-bound providers, and so storage read handles release at statement end. But
the mechanism is replaceable: the pooled session's `public` schema is now a lix-owned
`StatementSchemaProvider` (`RwLock<HashMap>`, identical registration semantics including
error-on-duplicate) whose recycle path is a single `clear()`. The enumeration fallback remains only
for a foreign schema provider reaching the pool.

**2. `SortExec` over a pinned point read is elided at plan construction (E38 item: part of 8.3 µs).**
A point read `WHERE pk = 'x' ORDER BY pk` (or single-value `IN`) streamed its at-most-one row
through a full `SortExec`. The scan now restores information the `Exact` filter pushdown had
removed from DataFusion's view: a fully-applied `col = literal` (or 1-element `IN`) pins that
output column to one value, so `SpecScanExec` declares it as a uniform constant in its equivalence
properties — exactly the fact `FilterExec` advertises when the same predicate stays above a scan.
DataFusion's own `EnforceSorting` rule then removes the sort during physical planning, so the
cached physical template never contains it and warm executions never sort. This is *not* a
runtime branch and not a ≤1-row cardinality claim: it is sound for any row count (a by-branch scan
returning N rows with the same pinned pk is still trivially ordered by it). Multi-value `IN`,
negated `IN`, `OR`, and `Inexact`/`Unsupported` filters pin nothing and keep their sort — covered
by tests in both directions.

**3. Warm cached reads skip the logical provider rebind (E38 item: 1.8 µs template rebind + part of session residue).**
On a physical-template cache hit, the statement's logical plan is provably never read — the cached
leaf-scan requests replan directly against the current snapshot and graft into the template. Yet
every warm statement still deep-walked the cached logical plan, resolved each table provider
async, and rebuilt the scan spine (`rebind_cached_read_plan`). Read plans now travel as
`RuntimeReadPlan::{Bound,Detached}`: statements owning a physical-cache key hand the detached
template over as-is, and rebinding happens inside the runtime only on the paths that actually plan
the logical tree (first execution per param set, template-fingerprint fallback). Provider
resolution goes through the same shared catalog list (`SessionState::schema_for_ref`), so the
rebound plan is identical when it is needed.

**Removed:** per-statement `table_names()` enumeration + per-name deregistration; the physical
`SortExec` on pinned point reads; the eager per-statement logical rebind on warm template hits.
**Not attempted (load-bearing):** per-statement provider *registration* (E38 item 4, 1.5 µs) is
what carries the statement's storage snapshot into the session — the warm template path consumes
those providers via `scan_with_args`, so hoisting them needs revision-keyed provider reuse, a
separate design. The sqlparser AST clone (0.85 µs) is also untouched.

## Layout invariants

1. **Single authority** — no new authority; a schema-provider implementation swap, an equivalence
   property, and a lazy rebind. No second writer or materialization.
2. **Commit canonicality** — untouched.
3. **Derived views are caches** — the planning caches remain revision/catalog-keyed caches;
   eviction/fallback paths preserved (template mismatch still falls back to the full planner).
4. **Atomic publication** — untouched.
5. **GC from refs** — untouched.
6. **SQL reads canonical records** — unchanged; providers still resolve per statement from the
   current snapshot.

## Verification

- New tests: `entity_point_read_order_by_pk_elides_physical_sort` (EXPLAIN has no `SortExec`,
  results identical) and `entity_multi_key_and_unpinned_order_by_keep_physical_sort` (multi-key IN
  and range-filtered ORDER BY keep their sort and correct order).
- Full CI-scope gate on ryzen-9950x-I (results below).
- A/B on ryzen-9950x-I, base = `claude-5-optimizations` head `d3f880ce`, cand = this branch.

### A/B — ryzen-9950x-I, RocksDB, warm sql_session hot lane, base SHA `d3f880ce`

Method: e38-run.sh's lane verbatim (TMPDIR on ext4, 5000 hot repeats per invocation, per-invocation
reseed, rotating arm order). SQLite arm is the byte-identical claim-3 binary, riding the headline
lane as reference + run-stability control (its 9-rep spread: 2.4%). Raw per-rep CSVs:
`~/claude5/e40-ab.csv`, `~/claude5/e40-upd9.csv`, `~/claude5/e40-null.csv` on the host.

| lane | rows | reps | base median | cand median | delta |
| **read_one_by_pk** | 10k | 9 | **29.12 µs** | **19.77 µs** | **−32.1%** (ranges disjoint: 28.6–29.5 vs 19.2–20.3) |
| read_one_by_pk | 1k | 3 | 29.09 µs | 19.97 µs | −31.4% |
| read_one_by_pk | 100k | 3 | 29.10 µs | 19.61 µs | −32.6% |
| read_many_by_pk (10) | 10k | 3 | 115.8 µs | 112.2 µs | −3.1% |
| update_one_by_pk | 10k | 9 | 271.6 µs | 280.7 µs | +3.3% — see null control below |
| read_all (warm) | 10k | 3 | 553.6 µs | 549.4 µs | −0.7% |
| read_all (warm) | 1k | 3 | 87.1 µs | 84.7 µs | −2.7% |

- **The curve is flat and the floor drops uniformly**: −9.3 µs at 1k, −9.4 µs at 10k, −9.5 µs at
  100k. This is a fixed per-statement term removed, exactly as the differential budget predicted —
  no scaling term added or removed elsewhere (read_all flat at both sizes).
- **SQLite ratio (this host): 31.1x → 21.1x** on the point read (SQLite median 936 ns, spread 2.4%).
- `read_all` here is the hot-lane metric that stops at `len()`; nothing in this change defers work
  past result construction (the recycle and rebind savings land inside the timed statement), so the
  consumed variant cannot diverge in the other direction.

### update_one +3.3% is the instrument, not the change — measured null control

The 9-rep escalation showed cand +3.3% on update_one with disjoint ranges, which looked real. The
null control resolves it: the cand worktree was reset to **the identical base commit `d3f880ce`**
and rebuilt, and that byte-equivalent-code binary was interleaved against the base binary, 5 reps:

| lane | base (base tree) | identical code (cand tree) | artifact |
|---|---:|---:|---:|
| update_one | 273.2 µs | 281.4 µs | **+3.0%, ranges disjoint** |
| read_one | 29.1 µs | 29.3 µs | +0.7% |

Two builds of the *same commit* from different worktree paths differ by +3.0% on the write lane —
binary-layout artifact (cargo fingerprints embed the worktree path; the two target trees lay code
out differently). The candidate's write delta net of the artifact is ≈ +0.3%, unresolvable. The
read lane's artifact is +0.7%, so the −32.1% read result stands far outside instrument error.
Mechanistically this is consistent: the write path shares only the schema-provider swap (a
per-statement `HashMap` vs dashmap, nanoseconds), and none of the three changes execute on it.

### Gate — full CI scope on ryzen-9950x-I (branch head `3c175da68`)

| step | result |
|---|---|
| `cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings` | exit 0 |
| `cargo test --profile test --workspace --exclude lix_benchmarks --all-features --no-fail-fast` | **3494 passed, 0 failed** |
| `cargo test --profile test -p lix_benchmarks --features storage-benches,slatedb,root-replay-trace --no-fail-fast` | **29 passed, 0 failed** |
| `cargo check -p lix --no-default-features` | exit 0 |
| `cargo check -p lix_js_sdk --target wasm32-unknown-unknown` | exit 0 |

Full logs: `~/claude5/e40-gate.log`, `~/claude5/e40-gate-tests.log`,
`~/claude5/e40-gate-bench-tests.log` on the host.

### Frame-pointer profile — the same instrument that found the items confirms their removal

Method identical to E38: dedicated `target-fp` build with
`RUSTFLAGS="-C force-frame-pointers=yes -C target-feature=+sse4.2,+pclmulqdq"` (env `RUSTFLAGS`
overrides the target rustflags, so the CRC32C features are re-added explicitly), `perf record
-F 2999 --call-graph fp` over 20 000 warm `read_one` repeats, samples bucketed by innermost owning
phase with type parameters stripped before matching (`~/claude5/e38-bucket2.awk`). FP arm measured
20.0 µs per operation (FP overhead vs the 19.8 µs A/B arm: 1.2%; E38's FP arm was 30.5 µs).

| E38 bucket (of 29.2 µs) | before | after (of 20.0 µs) |
|---|---:|---:|
| catalog `table_names()` at session recycle | 2.56 µs (8.8%) | **bucket absent (0 samples)** |
| `SortExec` (datafusion_sort) | inside the 8.33 µs stream-exec item | **bucket absent (0 samples)** |
| session recycle/drop residue | 1.14 µs (3.9%) | 0.33 µs (1.67%) |
| plan template rebind | 1.83 µs (6.3%) | 0.15 µs (0.73%) — the graft now rebuilds no sort node |
| DataFusion stream execution | 8.33 µs (28.5%) | ~4.2 µs (21.1% = df_execute 14.4% + collect 6.7%) |
| RocksDB | 4.10 µs (14.0%) | 3.53 µs (17.6%) — same absolute, larger share, as expected |

Profile artifact: `~/claude5/e40-lix-readone.perf`; full bucket table in `~/claude5/e40-fp.log`.

### What this means for claim 3

On this host the point read moves from **31.1x to 21.1x SQLite**. E38's remaining-incidentals
estimate said removing *everything* incidental lands near ~11x; what remains after this PR is the
read-session build (2.6 µs), provider registration (1.0 µs), the cooperative-stream execution
residue (~4.2 µs), statement plumbing incl. the sqlparser AST clone, and the architectural
versioning cost. Those are follow-up material (revision-keyed provider/session reuse; Arc-carried
parsed statements).

### DIFF
--- a/packages/lix/src/sql2/exec/datafusion.rs
+++ b/packages/lix/src/sql2/exec/datafusion.rs
@@ -77,7 +77,7 @@ pub(crate) const LIX_INSERT_COLUMN_OMITTED_METADATA_KEY: &str = "lix_insert_colu
 
 pub(crate) struct DataFusionLogicalPlan {
     pub(super) state: std::sync::Arc<SessionState>,
-    pub(super) plan: LogicalPlan,
+    pub(super) plan: crate::sql2::runtime::RuntimeReadPlan,
     pub(super) notices: Vec<LixNotice>,
     pub(super) json_predicate_params: BTreeSet<usize>,
     pub(super) expected_parameter_count: usize,
@@ -425,15 +425,25 @@ async fn create_logical_plan_in_session_from_parsed(
         && let Some((cache, catalog)) = &session.planning_environment
         && let Some(cached) = cache.read_plan(sql, params, catalog)
     {
-        let plan = rebind_cached_read_plan(session.context(), cached.plan.clone()).await?;
+        let physical_planning_cache = PhysicalReadPlanCacheKey::new(sql, params, catalog.clone())
+            .map(|key| (std::sync::Arc::clone(cache), key));
+        // With a physical-cache key the runtime rebinds scan providers lazily:
+        // a warm template execution never touches the logical plan, so eagerly
+        // resolving providers into it here would be pure per-statement waste.
+        let plan = if physical_planning_cache.is_some() {
+            crate::sql2::runtime::RuntimeReadPlan::Detached(cached.plan.clone())
+        } else {
+            crate::sql2::runtime::RuntimeReadPlan::Bound(
+                rebind_cached_read_plan(session.context(), cached.plan.clone()).await?,
+            )
+        };
         return Ok(SqlLogicalPlan::DataFusion(SqlDataFusionLogicalPlan {
             state: std::sync::Arc::clone(session.state()),
             plan,
             notices: Vec::new(),
             json_predicate_params: cached.json_predicate_params.clone(),
             expected_parameter_count: cached.expected_parameter_count,
-            physical_planning_cache: PhysicalReadPlanCacheKey::new(sql, params, catalog.clone())
-                .map(|key| (std::sync::Arc::clone(cache), key)),
+            physical_planning_cache,
         }));
     }
     bind_table_function_parameters(&mut statement, params)?;
@@ -473,7 +483,7 @@ async fn create_logical_plan_in_session_from_parsed(
 
     Ok(SqlLogicalPlan::DataFusion(SqlDataFusionLogicalPlan {
         state: std::sync::Arc::clone(session.state()),
-        plan,
+        plan: crate::sql2::runtime::RuntimeReadPlan::Bound(plan),
         notices: Vec::new(),
         json_predicate_params,
         expected_parameter_count,
@@ -659,7 +669,7 @@ async fn create_transaction_read_logical_plan_from_parsed(
     Ok((
         SqlLogicalPlan::DataFusion(SqlDataFusionLogicalPlan {
             state: std::sync::Arc::clone(session.state()),
-            plan,
+            plan: crate::sql2::runtime::RuntimeReadPlan::Bound(plan),
             notices: Vec::new(),
             json_predicate_params,
             expected_parameter_count,
@@ -1346,6 +1356,21 @@ fn validate_history_anchor_predicates_in_logical_plan(plan: &LogicalPlan) -> Res
 ///
 /// Mirrors `DataFrame::with_param_values`, which is exactly
 /// `LogicalPlan::with_param_values` plus a `SessionState` it does not need.
+fn bind_runtime_plan_param_values(
+    plan: crate::sql2::runtime::RuntimeReadPlan,
+    params: &[Value],
+) -> Result<crate::sql2::runtime::RuntimeReadPlan, LixError> {
+    use crate::sql2::runtime::RuntimeReadPlan;
+    Ok(match plan {
+        RuntimeReadPlan::Bound(plan) => {
+            RuntimeReadPlan::Bound(bind_plan_param_values(plan, params)?)
+        }
+        RuntimeReadPlan::Detached(plan) => {
+            RuntimeReadPlan::Detached(bind_plan_param_values(plan, params)?)
+        }
+    })
+}
+
 fn bind_plan_param_values(plan: LogicalPlan, params: &[Value]) -> Result<LogicalPlan, LixError> {
     if params.is_empty() {
         return Ok(plan);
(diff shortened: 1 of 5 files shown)


samuelstroschein:
Note on the base: the RUNBOOK directs drafts at `claude-5-optimizations`, but PR #1364 merged that branch into `main` and deleted it, so this draft targets `main` directly. The branch is cut from `d3f880ce` (the round-5 integration head at the time of the A/B and gate), which is an ancestor of current `main`; the two PRs merged since (#1397 storage GC sweep, #1398 RocksDB durability flag) do not touch the SQL layer files changed here. Coordinator: happy to rebase and re-gate if you want the head current before merging.

PR #1440, opened 2026-08-13
perf(hot-state): reclaim discharged tombstones at the checkpoint

samuelstroschein:
Serving-view tombstone compaction: a checkpoint now reclaims the tombstones it has just discharged.

## The term being removed

`ROW_SPACE` is a sparse serving overlay. A delete publishes a *tombstone* row there rather than removing the key, because the tombstone is what shadows a base row or a global-branch row. **Nothing ever removed one.** Measured (`hot_row_tombstone_probe`, phase 2): 999 tombstones from a 1000-row churn survive a checkpoint, a repository GC, a second checkpoint and a second GC. The serving layer carries an **O(deletes ever)** term where it should carry **O(live)**.

## Measured — the curve, not a point

`hot_row_tombstone_probe::checkpoint_compaction_scan_cost`, one live row at every size, deterministic (1 rep):

| n | entries after churn | entries after checkpoint | scan after churn | scan after checkpoint | compacted |
|---|---|---|---|---|---|
| 100 | 143 | **45** | 157 µs | 99 µs | 99 |
| 1 000 | 1 043 | **45** | 654 µs | 101 µs | 999 |
| 10 000 | 10 043 | **45** | 5 803 µs | 108 µs | 9 999 |

Row entries after the checkpoint are **flat at 45 across two orders of magnitude** while the pre-compaction count grows 70x, and the scan is flat at 99–108 µs while the un-compacted scan grows 37x. That is an asymptotic term removed, not a constant factor shaved. At n=1000 the point result is 1043 → 45 entries and 654 → 101 µs (6.5x).

## How

A checkpoint republishes the interval's dirty set with `reset_working_diff_baselines`, which is the one moment a tombstone's working-diff obligation is discharged. `hot_compaction_mask` runs there and marks what can go; `stage_hot_mutation_batch` already routes a `None` value range to a physical row delete, so removal adds **one disjunct to an existing branch** — no new staging function, no new space, no protocol bump.

Three conditions gate it, all conservative:

- **(a) the generation has no immutable base of any kind.** Packed current base, packed exclusive-schema base, sparse root base, certified entity batch — the four planes `has_schema_rows` composes besides `ROW_SPACE` itself. A checkpoint publishes no base of its own, so this is the ordinary state at a checkpoint. One key-only existence probe per plane per publication, not per identity.
- **(b) the global branch has no rows in the tombstone's schema.** Checked against what is durable (`has_schema_rows`, which is base-complete, so a global row whose authority is a base still counts) **and** against what this transaction publishes on global. Verdicts are cached per distinct schema key. Vacuous when the branch *is* global.
- **(c) the row is provably clean at removal.** `reset_working_diff_baselines` forces `(Clean, newly_dirty=false)` for every tracked delta, so this is free — and it is asserted rather than assumed: a compacted row that is not clean fails the publication.

**There is no fourth condition.** `DIFF_SPACE`, not `ROW_SPACE`, is the working diff's authority. The probe's `near_miss` arm measures that and keeps an inverted assertion so a change that reintroduces the dependency fails loudly.

## Three things in the brief that did not survive contact with the code

**1. Gate (b)'s transaction half, as approved, would have made compaction permanently inert.** The approved design paired the per-schema check with a transaction-wide bool, `tracked_roots.iter().any(|r| r.branch_id == GLOBAL_BRANCH_ID)`. But `checkpoint_stage_row` (`checkpoint.rs:56`) writes the `lix_checkpoint` row with `branch_id: GLOBAL_BRANCH_ID`, so **every checkpoint transaction carries a global tracked root** — and compaction only ever runs at a checkpoint. The bool is `true` every time it is read. This is exactly the failure the brief names ("a wholesale global-branch test blocks every checkpoint, and compaction is silently inert"), arriving through the mechanism proposed to avoid it.

The fix keeps the ordering guarantee and applies the brief's own rule to both halves: the transaction half is a **set of schema keys**, `global_branch_schema_keys`, derived from the transaction's prepared inputs (prepared state rows, engine rows, global roots' selected change batches, the synthesized branch-ref row) rather than from the write set. That is order-independent for the same reason the bool was — it never consults staging order — so it closes the `visit_tracked_root_parent_first` hole completely, while leaving `lix_checkpoint` from blocking an unrelated schema. A global root carrying a lifecycle snapshot is not enumerated and forces the disabling `None`.

**2. Gate (a) is four planes, not two.** The brief and phase 4's premise check `packed_bases` and `root_bases`. The serving view also composes `PACKED_CURRENT_EXCLUSIVE_SCHEMA_BASE_SPACE` and certified entity batches; `has_schema_rows` — the engine's own answer to "does this generation hold rows in this schema" — consults both. A tombstone shadowing either would have resurrected a row. Gate (a) probes all four.

**3. The prize fixture measures a case this change does not reclaim.** The 1043 → 44 premise was measured on *churn, then first-ever checkpoint*. A checkpoint republishes the interval's **dirty set**, and a delete only joins that set while a checkpoint interval is open. Measured on that exact fixture: the route runs and reports `routes=1 offered=2` — the deletes are simply not in the publication. With a baseline checkpoint before the churn (the steady state, and what the corrected fixtures use) the full win lands. **Deletes made before a repository's first checkpoint are a separate, larger problem this change does not claim.** This is stated in a comment on the test that found it.

## Engagement — counted inside the route, before any timing

`COMPACTED_TOMBSTONE_{ROUTES,OFFERED,CANDIDATES,COMPACTED}`, incremented inside `hot_compaction_mask` itself, gated `#[cfg(any(test, feature = "storage-benches"))]`, re-exported with `#[allow(unused_imports)]` (the counters are read only by `#[cfg(test)]` probes, so a features-on *library* build has no reader and only `clippy --all-targets` sees it — the omission that failed #1432's gate).

`ROUTES`/`OFFERED` sit above `CANDIDATES` deliberately: "the checkpoint offered no tombstone" and "the checkpoint never reached the mask" have identical readings at `CANDIDATES`, and that distinction is what found item 3 above.

## Tests

Non-`#[ignore]`d, both verdicts:

- `checkpoint_compacts_discharged_branch_tombstones` — the hit, end to end through SQL: 343 → 45 entries, 299 tombstones → 0, `candidates=299 compacted=299`, answers unchanged after a **cold reopen**, working diff still 0.
- `checkpoint_compaction_keeps_only_globally_shadowed_tombstones` — gate (b), both verdicts in one publication: a global-branch row and a branch tombstone in the same schema (tombstone survives) alongside a control in a schema global has no rows in (reclaimed). Deterministic, at the exact layer.
- `recreated_identity_inherits_created_at_after_engine_compaction` — the inverted h7 pin against **real** compaction rather than the simulated `drop_all_tombstones`: `assert_eq!` that a re-insert over a genuinely compacted identity inherits its original `created_at`, with `canonical_hits > 0` proving the canonical sourcing actually fired, plus a successful `rebuild_tracked_state_for_branch`. This is the failure mode with no other guard — a wrong timestamp is written silently and only surfaces on a later rebuild.
- `checkpoint_compaction_scan_cost` — the `#[ignore]`d measurement above. **Appended as PHASE 10**; phases 1–9 are untouched.

Counter assertions are thresholds scaled to the fixture, never exact values: `storage_bench` counters are process-global.

Phase 9's `recreated_identity_created_at_after_compaction` now finds nothing to remove, because real compaction beat its simulation to the tombstone. Its `removed > 0` assertion is replaced with `after.tombstones == 0` — the property it was actually protecting — with a comment saying why.

**The five global-shadow regression tests are unmodified and green.** `hot_state/visibility.rs` and `hot_state/context.rs`, where four of the five live, are not touched by this diff at all.

## Where the flag lives

`transaction_global_schema_keys` is a **field on `HotStateWriter`**, not a parameter threaded through the staging wrappers, because it is a property of the *transaction* rather than of a call. It is set through a builder (`with_transaction_global_schema_keys`) rather than the factory signature: **1 field + 1 default + 1 call site**, against ~11 sites for a factory-signature change, in the most contended function in the tree.

**Polarity.** `None` is the factory default and **disables** compaction. A construction site that cannot enumerate what its transaction stages on global is therefore safe by omission rather than by vigilance — the inversion that reads as obviously correct in review is not reachable here.

## Layout invariants

1. **Single authority** — no. Nothing gains a second authority; a derived view loses dead weight. Canonical tracked state is untouched, and `TrackedStateDiffEntry.after` keeps its raw tombstone (its doc-comment justifies that as what lets merge apply deletes without reloading the source root).
2. **Commit canonicality** — unaffected. No commit record, parent link or state root changes.
3. **Derived views are caches** — this is the point of the change: `ROW_SPACE` is a rebuildable serving view, and a tombstone that shadows nothing is cache weight. Undo/redo replay from canonical state, not from the serving row (phase 6 measures that).
4. **Atomic publication** — yes. The removals are staged into the checkpoint's own `StorageWriteSet`, so they publish in the same atomic write set as the checkpoint that discharged them. No reader can observe a half-compacted generation.
5. **GC from refs** — unaffected. No retention metadata added.
6. **SQL reads canonical records** — unaffected.

## Breaking status

- **Public API** — non-breaking. No change to `packages/lix/src/lib.rs` exports, the SQL surface, or the JS SDK surface.
- **Storage format / physical layout** — non-breaking. No new persisted field, no new space, no `REPOSITORY_PROTOCOL_VALUE` bump. Compaction removes keys an existing reader already treats as absent.
- **Storage adapter API** — non-breaking. Nothing added, changed or removed on the `StorageAdapter` trait; both shipping adapters and the conformance suite are untouched.

## Gate

Full CI scope, **re-derived from this sha's own `ci.yml`** (the `lix_e2e` feature list is extracted by the gate script itself, not carried forward — it gained `plugin-tests,server-protocol` in the integration, and a stale command would have silently skipped three targets). Run on `hetzner-cpx62-III`.

```
git rev-parse HEAD        7b42d673a02a50600960930478bae1be49222622
git status --porcelain    (empty)
CI_SCOPE_CONFIRMED_AT=7b42d673a02a50600960930478bae1be49222622
EXECUTED_TARGETS test1=3426 test2=172
```

```
E2E_FEATURES_FROM_CI_YML=sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace

CLIPPY_EXIT=0      cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings
NODEFAULT_EXIT=0   cargo check -p lix --no-default-features
WASM_EXIT=0        cargo check -p lix_js_sdk --target wasm32-unknown-unknown
TEST1_EXIT=0       cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast
TEST2_EXIT=0       cargo test --profile test -p lix_e2e --features <from that sha> --no-fail-fast
CANARY_EXIT=0      slatedb_history_blob_survives_until_final_root_release, by name, 1 passed
```

**Counts accounted for.** Against the merged tree's reference (test1 3423 passed, test2 172 passed / 27 targets): test2 matches exactly; test1 is **+3**, which is exactly the three non-`#[ignore]`d tests this PR adds — all three confirmed `... ok` by name in the log. This diff adds no test *target*, only tests inside existing ones. `TARGET_COUNTS test1=38 test2=27`, of which 12 of test1's are doc-test targets.

**The newly reached surface passed, and was not vacuous:** `plugin_gc` → `ok. 2 passed`, `server_protocol_durable_idempotency` → `ok. 2 passed`. Non-zero on both, which is the point of checking.

`GATE_HEAD_BEFORE == GATE_HEAD_AFTER == 7b42d673a` — the worktree is shared across agents on that host, so the gate asserts its own HEAD on both sides of the run. Every log is truncated before the run, so a previous run's result cannot be read as this one's.

## Rebase onto the integration — verified by content

Rebased `be813172a` → **`ae6ef49d0`** (the `main` integration: plugin runtime boundary hard cut). `git` reported zero conflicts, which is not the evidence — this is:

- **My added/removed lines are byte-identical to the pre-rebase gated diff.** Diffed content lines of both patches: no difference.
- **The `main` side survived in my files**: `0` stale `crate::wasm::` / `crate::plugin_wire` / `crate::plugin_layout` references in either `transaction/commit.rs` or `hot_state/tracked_head/hot.rs`, with `8` and `14`+`15` references to the new `crate::plugin::runtime` / `crate::plugin::wire` paths respectively.
- **Every hunk containing a deletion was read end to end** — the four in-place edits (the `.enumerate()` on the values loop, the gate (c) check plus the `physically_deletes() || compacted_row` push, the writer factory, and the counter re-export). No function was spliced; the gate's compile is the independent confirmation.
- **The sealed-owner-boundary assertion in `code_structure.rs` is not triggered:** this diff introduces **no** `plugin::` import at all.

### DIFF
--- a/packages/lix/src/hot_row_tombstone_probe.rs
+++ b/packages/lix/src/hot_row_tombstone_probe.rs
@@ -1380,7 +1380,16 @@ async fn recreated_identity_created_at_after_compaction() {
          entries {}->{} tombstones {}->{}",
         before.entries, after.entries, before.tombstones, after.tombstones
     );
-    assert!(removed > 0, "the delete must have left a tombstone to remove");
+    // The engine now compacts a discharged tombstone at the checkpoint itself
+    // (`hot_compaction_mask`), so by this point there is usually nothing left
+    // for the simulation to remove. Either state is a valid entry into the
+    // question this arm asks — what a re-insert over a *missing* tombstone
+    // gets — and phase 10 pins the engine-driven route with its own test. What
+    // must hold here is that no tombstone survives into the re-insert.
+    assert_eq!(
+        after.tombstones, 0,
+        "the identity must carry no tombstone into the re-insert (removed={removed})"
+    );
 
     let session = reopen_session(&storage).await;
     let hits_before = crate::hot_state::BROAD_CANONICAL_CREATED_AT_HITS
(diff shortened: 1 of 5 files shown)


samuelstroschein:
## Base movement since the gate — no re-gate

`claude-6-optimizations` moved `be813172a` → `910608a07` after this PR was gated. That is **#1437 alone** (`perf(history): pin file_ids on the file-history descriptor/blob route`), not the pending `main` integration.

- **Gated sha:** `2afa2a007e13baa27c85687c891ffb0e5df1b1cb` (PR head, unchanged).
- **Landed after it:** #1437, touching `sql2/exec/datafusion.rs`, `sql2/providers/file_history.rs`, `e2e/tests/history_scaling_probe.rs`.
- **Why it cannot affect this result:** zero file overlap with this diff (`hot_state/*`, `transaction/commit.rs`, `hot_row_tombstone_probe.rs`), and #1437 sits in the SQL provider/exec layer, which reads *from* the hot serving view rather than underpinning it. Nothing in it can change what a checkpoint stages into `ROW_SPACE`.

Per the runbook's "do not re-gate on every base movement" rule this is stated rather than re-run, which keeps the claim falsifiable at no machine cost.

**This PR is deliberately still based on `be813172a`** so that its head remains the sha the gate evidence is attached to. It will be rebased and **fully re-gated** on the `main` → `claude-6-optimizations` integration when that lands — that one has real overlap (`transaction/commit.rs` and `hot_state/tracked_head/hot.rs` on both sides, plus a plugin-runtime hard cut), so it needs the full run, not a note.

PR #1453, opened 2026-08-13
perf(sql2): decide entity row predicates without a snapshot DOM

samuelstroschein:
## What this is

A profile of a 200k-row entity scan (mimalloc, frame pointers, self time gated to the read
subtree through `execute_read_statement_with_store`) attributed **38.2% of the read subtree to
`apply_entity_batch_filters`** — and roughly 70% of *that* to building and tearing down a
`serde_json::Value`:

| leaf inside `apply_entity_batch_filters` | share of that subtree |
|---|---|
| `ValueVisitor::visit_map` | 26.6% |
| `BTreeMap<String, Value>::insert` | 15.6% |
| `bytes::shared_drop` | 14.6% |
| `from_trait::<StrRead, Value>` | 5.8% |
| `mi_free` | 6.2% |
| `parse_number` / `parse_str` / `skip_to_escape` | 6.7% |

Evaluating the predicate itself — `EntityRowFilter::matches_snapshot` — was **0.56% of the whole
profile**.

## The mechanism

`apply_entity_batch_filters` parsed each candidate row's entire `snapshot_content` into a
`serde_json::Value` **in order to decide the predicate**, and did so for every **scanned** row, not
every returned row. `EntityRowFilter` is only `ColumnEq` / `ColumnIn` / `And` / `Or` over *named
top-level* columns, so a scan that returns two columns was materialising a `BTreeMap` of every
field of every row to read one of them.

The parse is the predicate's **input**, so it cannot be moved behind the filter — "filter earlier
so we parse fewer rows" does not apply on this route. What it can be is **narrower**.

## The change

- `EntityRowFilter::collect_filter_columns` names every top-level column the predicate tree reads.
- `parse_snapshot_filter_columns` streams the document with a `serde` visitor and materialises only
  those columns, skipping every other value with `IgnoredAny`.
- `FilteredEntityBatch::parsed_snapshots` — the side column that existed only so projection could
  reuse the filter's DOM — is removed, because there is no longer a DOM to reuse. Filtered batches
  now reach the same projection routes unfiltered ones already use, including the streaming
  `EntityProjectionDecoder`, which builds Arrow with no DOM at all.

Full snapshot parses per scan: **one per scanned row → at most one per surviving row, and zero on
the tracked raw-projection route.**

**No format change.** Nothing about what is stored or encoded moves; the diff touches one file and
no write-path or encoding symbol.

## Correctness

A partial parse that silently drops a field returns wrong rows, so the rejection cases are asserted
as hard as the acceptance cases. `parse_snapshot_filter_columns` keeps `serde_json::from_str`
semantics on every axis that decides a row:

- the **whole document is still validated** (`Deserializer::end()` is still called), so malformed
  JSON and trailing bytes are still rejected at the same place — including damage inside a value
  the predicate skips over;
- **duplicate keys are still last-wins** (a first-wins reader returns a different row);
- **escaped keys and escaped values are still unescaped before comparison** (`body` still
  matches the column `body`);
- a **non-object snapshot** still yields a value from which no column can be read, matching what
  `JsonValue::get` did on the original.

Three new tests in `packages/lix/src/sql2/providers/entity.rs`:

- `predicate_column_parse_matches_the_full_snapshot_parse` — differential against
  `parse_snapshot_value` over 15 acceptance fixtures (duplicate keys, escaped key, escaped value,
  nested decoy key, non-object roots, empty object) and 5 rejection fixtures, asserting each
  rejection fixture is *also* rejected by the full parser so the fixture cannot silently rot.
- `filtered_entity_scan_never_builds_a_candidate_snapshot_dom` — the engagement check. It asserts
  the **new** counter `ENTITY_SNAPSHOT_FILTER_PARSE_COUNT` saw both candidates, then that
  `ENTITY_SNAPSHOT_PARSE_COUNT` is 0. Counting only full parses would have made "the fix worked"
  and "the filter route never ran" read identically — both zero.
- `collect_filter_columns_walks_the_whole_predicate_tree` — a column missed here silently drops the
  predicate that reads it.

## Gate

See the comment below. Base: `claude-7-optimizations`. **Do not merge — coordinator merges.**

### DIFF
--- a/packages/lix/src/sql2/providers/entity.rs
+++ b/packages/lix/src/sql2/providers/entity.rs
@@ -17,6 +17,7 @@ use datafusion::logical_expr::{BinaryExpr, Expr, Operator, TableProviderFilterPu
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::prelude::SessionContext;
 use futures_util::FutureExt;
+use serde::de::{DeserializeSeed, IgnoredAny, MapAccess, SeqAccess, Visitor};
 use serde_json::Value as JsonValue;
 
 use crate::branch::BranchRefReader;
@@ -395,12 +396,11 @@ impl EntitySpec {
             .scan_batch(&request)
             .await
             .map_err(lix_error_to_datafusion_error)?;
-        let batch = entity_record_batch_with_parsed(
+        let batch = entity_record_batch(
             &self.spec,
             Arc::clone(&self.schema),
             &rows,
             EntityBatchProjection::for_request(&request),
-            None,
         )?;
         let mut post_rows = BTreeMap::new();
         for row_index in 0..batch.num_rows() {
@@ -481,13 +481,7 @@ impl EntitySpec {
                     .map_err(lix_error_to_datafusion_error)?;
                 let filtered = apply_entity_batch_filters(rows, &row_filters)?;
                 capture_entity_update_snapshots(&filtered.rows, &update_snapshots)?;
-                entity_record_batch_with_parsed(
-                    &spec,
-                    schema,
-                    &filtered.rows,
-                    batch_projection,
-                    filtered.parsed_snapshots.as_deref(),
-                )
+                entity_record_batch(&spec, schema, &filtered.rows, batch_projection)
             },
         );
         let spec = Arc::clone(&self.spec);
@@ -774,13 +768,7 @@ impl TableSpec for EntitySpec {
                     // predicate without an indexed access path pays for.
                     record_rows_examined(rows.len());
                     let filtered = apply_entity_batch_filters(rows, &row_filters)?;
-                    entity_record_batch_with_parsed(
-                        &spec,
-                        schema,
-                        &filtered.rows,
-                        batch_projection,
-                        filtered.parsed_snapshots.as_deref(),
-                    )
+                    entity_record_batch(&spec, schema, &filtered.rows, batch_projection)
                 },
             ),
         })
@@ -832,13 +820,7 @@ impl TableSpec for EntitySpec {
                     .await
                     .map_err(lix_error_to_datafusion_error)?;
                 let filtered = apply_entity_batch_filters(rows, &row_filters)?;
-                entity_record_batch_with_parsed(
-                    &spec,
-                    schema,
-                    &filtered.rows,
-                    batch_projection,
-                    filtered.parsed_snapshots.as_deref(),
-                )
+                entity_record_batch(&spec, schema, &filtered.rows, batch_projection)
             },
         );
         let spec = Arc::clone(&self.spec);
@@ -2501,6 +2483,23 @@ impl EntityRowFilter {
         }
     }
 
+    /// Names every top-level snapshot column this predicate reads.
+    ///
+    /// Row predicates are `ColumnEq` / `ColumnIn` over visible top-level
+    /// columns, so the set is small and known before the scan decodes a
+    /// single row.
+    fn collect_filter_columns<'a>(&'a self, out: &mut BTreeSet<&'a str>) {
+        match self {
+            Self::ColumnEq { column, .. } | Self::ColumnIn { column, .. } => {
+                out.insert(column.as_str());
+            }
+            Self::And(left, right) | Self::Or(left, right) => {
+                left.collect_filter_columns(out);
+                right.collect_filter_columns(out);
+            }
+        }
+    }
+
     fn matches_snapshot(&self, snapshot: Option<&JsonValue>, schema_key: &str) -> Result<bool> {
         match self {
             Self::ColumnEq {
@@ -2894,18 +2893,23 @@ fn apply_entity_batch_filters(
     filters: &[EntityRowFilter],
 ) -> Result<FilteredEntityBatch> {
     if filters.is_empty() {
-        return Ok(FilteredEntityBatch {
-            rows,
-            parsed_snapshots: None,
-        });
+        return Ok(FilteredEntityBatch { rows });
+    }
+    let mut filter_columns = BTreeSet::new();
+    for filter in filters {
+        filter.collect_filter_columns(&mut filter_columns);
     }
     let mut filtered = MaterializedHotStateBatchBuilder::with_capacity(rows.len());
-    let mut parsed_snapshots = Vec::with_capacity(rows.len());
     for row in rows.iter() {
         let Some(snapshot_content) = row.snapshot_content().map(AsRef::<str>::as_ref) else {
             continue;
         };
-        let snapshot = parse_snapshot_value(snapshot_content).map_err(|error| {
+        // The predicate reads a handful of named columns. Materializing the
+        // whole snapshot map here charged every *scanned* row for a DOM the
+        // predicate never looked at; the projection below decodes the
+        // *surviving* rows on its own streaming path.
+        let snapshot = parse_snapshot_filter_columns(snapshot_content, &filter_columns)
+            .map_err(|error| {
             DataFusionError::External(Box::new(LixError::new(
                 LixError::CODE_INTERNAL_ERROR,
                 format!(
@@ -2914,7 +2918,7 @@ fn apply_entity_batch_filters(
                     row.entity_pk()
                 ),
             )))
-        })?;
+            })?;
         let mut matches = true;
         for filter in filters {
             if !filter.matches_snapshot(Some(&snapshot), row.schema_key())? {
@@ -2924,21 +2928,15 @@ fn apply_entity_batch_filters(
         }
         if matches {
             filtered.push_ref(row, None);
-            parsed_snapshots.push(Some(snapshot));
         }
     }
     Ok(FilteredEntityBatch {
         rows: filtered.finish(),
-        parsed_snapshots: Some(parsed_snapshots),
     })
 }
 
 struct FilteredEntityBatch {
     rows: MaterializedHotStateBatch,
-    /// Parsed snapshots retained only when row predicates already needed a
-    /// DOM. Projection consumes this side column instead of parsing winners a
-    /// second time.
-    parsed_snapshots: Option<Vec<Option<JsonValue>>>,
 }
 
 fn entity_hot_state_scan_request(
@@ -3076,31 +3074,16 @@ impl EntityBatchProjection {
     }
 }
 
-#[cfg(test)]
 fn entity_record_batch(
     spec: &EntitySurfaceSpec,
     schema: SchemaRef,
     rows: &MaterializedHotStateBatch,
     projection: EntityBatchProjection,
 ) -> Result<RecordBatch> {
-    entity_record_batch_with_parsed(spec, schema, rows, projection, None)
-}
-
-fn entity_record_batch_with_parsed(
-    spec: &EntitySurfaceSpec,
-    schema: SchemaRef,
-    rows: &MaterializedHotStateBatch,
-    projection: EntityBatchProjection,
-    parsed_snapshots: Option<&[Option<JsonValue>]>,
-) -> Result<RecordBatch> {
     if schema.fields().is_empty() {
         let options = RecordBatchOptions::new().with_row_count(Some(rows.len()));
         return RecordBatch::try_new_with_options(schema, vec![], &options)
             .map_err(DataFusionError::from);
-    }
-    if let Some(parsed_snapshots) = parsed_snapshots {
-        debug_assert_eq!(parsed_snapshots.len(), rows.len());
-        return entity_record_batch_from_parsed_snapshots(spec, schema, rows, parsed_snapshots);
     }
 
     match projection {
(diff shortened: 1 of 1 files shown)


samuelstroschein:
## Gate — green

Re-derived from **this PR's own sha's** `.github/workflows/ci.yml`. That file has **no
`--no-default-features` step and no `wasm32-unknown-unknown` cargo step** in the test job, and the
`lix_e2e` feature list is `sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace`.

```
git rev-parse HEAD        049392ead34ccbad3fe9c3d0bf36422d77335091   (before AND after the build)
git status --porcelain    <empty>                                    (before AND after the build)
CI_SCOPE_CONFIRMED_AT=049392ead34ccbad3fe9c3d0bf36422d77335091
CLIPPY_EXIT=0  TEST1_EXIT=0  TEST2_EXIT=0
EXECUTED_TARGETS test1=38 test2=27
PASSED           test1=3432 test2=172
NEW_TEST_PRESENT=1 filtered_entity_scan_never_builds_a_candidate_snapshot_dom
NEW_TEST_PRESENT=1 predicate_column_parse_matches_the_full_snapshot_parse
NEW_TEST_PRESENT=1 collect_filter_columns_walks_the_whole_predicate_tree
```

Reference is 3430 / 38 targets. This PR replaces one test with three, so **3432** is the expected
count — a count that had not moved would have meant the tests did not run.

## Fixture — state it, never inherit it

| | |
|---|---|
| host class | `ryzen-9950x-V`, AMD Ryzen 9 9950X 16-core |
| storage backend | RocksDB |
| rows | T = 200 000, one user-registered schema `olap_row` (5 columns: 2 string, 1 int, 1 float, 1 bool) |
| seed | `begin_transaction` + multi-row `INSERT ... VALUES` at 4096 rows/statement |
| checkpoint cadence | 1 checkpoint after seed |
| delete history | **none — freshly seeded, zero tombstones**, so this is not measuring tombstone decode |
| packed base | **not published.** This seed shape is not the certified single-`execute_batch` route, so this measures the **unpacked hot row plane** — which is the path ordinary workloads take |
| allocator | **mimalloc** as `#[global_allocator]` (bench line 13) — confirmed, not assumed |
| build | both arms from the same bench source (`md5` identical), env `RUSTFLAGS="-C force-frame-pointers=yes -C target-feature=+sse4.2,+pclmulqdq"`, separate `target-fp-{base,cand}`. No `--config target.<triple>.rustflags` and no `-fuse-ld=mold` |
| arms differ by | `packages/lix/src/sql2/providers/entity.rs` only, 1 file |

## Deterministic count (1 rep — it is a count, not a rate)

**Full snapshot DOM parses per scan: one per *scanned* row → zero in the filter step, and at most
one per *surviving* row in projection** (zero on the tracked raw-projection route). Asserted by
`filtered_entity_scan_never_builds_a_candidate_snapshot_dom` against two counters, one of which is
new so that "the fix worked" and "the route never ran" cannot both read as zero.

## Profile, before and after — same instrument, same binaries as the timing

`perf --call-graph fp`, gated to the read subtree through `execute_read_statement_with_store`
(at T=200k the seed phase is a large share of samples; ungated normalisation manufactures a chain
pointing at the *insert* path). Shares are of **gated** stacks. `range` N=1, 8 s window.

| | base | cand |
|---|---|---|
| gated read stacks | 9 190 | 8 993 |
| `apply_entity_batch_filters` | **37.62%** | **22.77%** |
| any `serde_json` frame | **31.76%** | **7.67%** |
| `serde_json::Value` DOM construction (`ValueVisitor` + `BTreeMap<String, Value>::insert`) | **29.49%** | **2.55%** |

Top leaves inside `apply_entity_batch_filters`:

| base | | cand | |
|---|---|---|---|
| `ValueVisitor::visit_map` | 24.65% | `bytes::shared_drop` | 25.63% |
| `BTreeMap<String, Value>::insert` | 15.85% | `apply_entity_batch_filters` (the skip loop) | 25.29% |
| `bytes::shared_drop` | 13.91% | `StrRead::parse_str` | 7.62% |
| `mi_free` | 8.88% | `_mi_page_malloc_zero` | 7.42% |
| `from_trait::<StrRead, Value>` | 5.84% | `SliceRead::skip_to_escape` | 6.30% |
| `deserialize_any::<ValueVisitor>` | 4.28% | `BTreeMap<String, Value>::insert` | 4.79% |

The claim was "JSON parsing fell", so the evidence is those symbols shrinking: **DOM construction
29.49% → 2.55%**, and `from_trait::<StrRead, Value>` leaves the top leaves entirely. The residual
2.55% is intended — it is the one full parse the *surviving* row still gets on the projection
route, plus the single `Map::insert` the partial parser does for the predicate's own column.

The largest remaining leaf, `bytes::shared_drop` at 25.63%, is the separate `bytes::Bytes`
refcounting target and is untouched here.

## Timing A/B — 5 reps, interleaved with arm-order rotation

Each rep is a fresh process on a fresh RocksDB directory. Times are µs per query.

| shape | N | base median / p95 / raw | cand median / p95 / raw | result |
|---|---|---|---|---|
| `range` | 1 | 141 025 / 141 996 <br>[140089 140969 141025 141235 141996] | 125 349 / 125 462 <br>[125090 125312 125349 125420 125462] | **−11.1%, disjoint ranges** |
| `range` | 100 | 191 606 / 192 130 | 194 368 / 196 504 | overlapping ranges |
| `range` | 10 000 | 194 283 / 194 413 | 194 413 / 197 023 | overlapping ranges |
| `point` | 1 | 28.4 / 29.8 | 28.7 / 28.9 | overlapping ranges |

**Where the change engages, and where it provably cannot.** `range` N=1 is the target: DataFusion
folds `ordinal BETWEEN k AND k` into `ordinal = k`, which becomes an `EntityRowFilter::ColumnEq` on
a **non-key** column, so the provider walks all 200 000 rows and decides each one with a snapshot
parse. That is the only cell that moves, and it moves with **disjoint ranges** — well clear of the
~5% band below which non-overlap is not evidence on this harness.

The other three cells are controls, and all three come back overlapping as designed:

- **`point` N=1** is the intended null control: `id` is the declared primary key, so the predicate
  routes through the exact-identity access path, `row_filters` is empty, and
  `apply_entity_batch_filters` returns before any changed line.
- **`range` N=100 / N=10 000** turn out to be controls too, and this is worth recording:
  `EntityRowFilter` has **no comparison variants at all** (`ColumnEq`, `ColumnIn`, `And`, `Or`), so
  a genuine `BETWEEN lo AND hi` with `lo != hi` produces **no row filter**. Those scans never enter
  the changed function. Their being flat is a positive result, not a null one.

**Answers are byte-identical across both arms and all reps.** The bench folds a hash over every
cell of every returned row; across 2 arms x 5 reps, every `(shape, N)` yields exactly one distinct
`(fold, row_count)`.

## Known sibling, deliberately not in this PR

`entity_columnar_overlay_batches` (same file) has the same defect on the columnar-overlay route:
`parse_snapshot_value` per candidate row purely to decide `row_filters`. It is a drop-in for the
same call. It is **not** in this PR because that route sits on the columnar tier, which is
unreachable in ordinary workloads, so it would carry a re-gate cost for no measurable gain and a
non-zero risk. Flagging it rather than silently leaving it.

samuelstroschein:
## Collision check with the `ColumnRange` range-pushdown work — **already compiler-enforced, no edit needed**

The hazard is real and worth naming: `parse_snapshot_filter_columns` materialises **only** the
columns `collect_filter_columns` enumerates. A predicate variant whose column is not enumerated
there evaluates against **absent data** — wrong rows, silently, with no compile error and no panic.

`collect_filter_columns` matches exhaustively on `EntityRowFilter` with **no catch-all arm**, so
adding `ColumnRange` cannot compile until someone handles it.

**Proved mechanically, not by grep** — grep is what produced the false alarms in this campaign, and
there *are* two `_ => None` arms inside `impl EntityRowFilter` that a grep flags: both are on a
tuple of `Option<bool>` inside `may_match_group`, not on `Self`. So the grep answer is misleading
in exactly the direction that matters.

Instead, a probe variant standing in for `ColumnRange` was added to the enum and
`cargo check -p lix --all-features --all-targets` was allowed to enumerate the sites. It failed
with `EXIT=101` and **four** `E0004: non-exhaustive patterns` errors, one per `match self`:

```
packages/lix/src/sql2/providers/entity.rs:1872  fn declared_column_membership
packages/lix/src/sql2/providers/entity.rs:2443  fn may_match_group
packages/lix/src/sql2/providers/entity.rs:2492  fn collect_filter_columns   <-- this PR's site
packages/lix/src/sql2/providers/entity.rs:2504  fn matches_snapshot
```

(line numbers given for the reverted tree; the probe shifted them +8 while it was applied)

`collect_filter_columns` is in the list. The probe was reverted immediately;
`git status --porcelain` is empty and `HEAD` is still `049392ead34ccbad3fe9c3d0bf36422d77335091`.

**So the pair composes safely.** For whoever lands `ColumnRange`, the contract at that site is:
*enumerate every top-level column the variant reads in `matches_snapshot`.* For a range that is the
one `column` field — the same shape `ColumnEq` and `ColumnIn` already have, so the new arm folds
into the existing `Self::ColumnEq { column, .. } | Self::ColumnIn { column, .. }` arm.

The composition is worth having: once `ColumnRange` exists, genuine `BETWEEN lo AND hi` predicates
enter `apply_entity_batch_filters` for the first time, which is precisely the case this PR measured
as **out of reach today** (the flat `range` N=100 / N=10 000 cells above). The two changes cover
disjoint halves of the same route.

## `cargo check -p lix --no-default-features`

```
NO_DEFAULT_FEATURES_EXIT=0
```

Clean. One `dead_code` warning is emitted, and it is **not** from this PR:
`CommitTouchedScopeDigest::absent` in `packages/lix/src/changelog/scope_digest.rs`. This PR's diff
is one file (`sql2/providers/entity.rs`), that file contains **zero** references to
`CommitTouchedScopeDigest` or `scope_digest`, and the commit touches no other module — so it cannot
have been this PR that stranded it. Pre-existing and unrelated.

This PR also touches **no census or `storage_bench` call sites**: the only counters it adds are the
two `#[cfg(test)]` thread-locals in `entity.rs`, which are absent from any non-test build, which is
why the feature-gated surface is unaffected.

samuelstroschein:
## Union of #1453 + #1454 measured on the one fixture where both halves are live

**They are additive, and the union is disjoint from both singles.** Union kept **local and unpushed**
(`514fc643b`); this comment is measurement only.

### Why this fixture and not the e53 scan baseline

The earlier 2x2 on `expb_scan_baseline` could not test this: `filterable_column_name` returns `None`
for `EntityColumnType::Json`, so `WHERE value = lix_json($1)` on `lix_key_value` never becomes an
`EntityRowFilter`, and `apply_entity_batch_filters` never runs — **zero frames in an 18 508-line
profile script**, with a working-grep control (231 `providers::entity` frames) proving the profile
was not simply empty. That harness's null control moved **+5.2% disjoint**, putting its floor at
≥5%.

`oltp_curve` `range` N=1 is different: DataFusion folds `ordinal BETWEEN k AND k` to `ordinal = k`
on an **integer, non-key** column, which is a real `ColumnEq` reaching the provider.

### Artifact band, established for THIS harness before quoting any delta

Measured on the three cells where #1453 is structurally inert (no `EntityRowFilter` is produced, so
the function returns before any changed line):

| null cell | B vs A |
|---|---|
| `point` N=1 | +2.9% |
| `range` N=100 | +1.5% |
| `range` N=10 000 | +1.0% |

**Band = 2.9%.** Every claim below is reported against that, not against e53's ≥5%.

### The 2x2

Four dedicated target dirs; identity asserted **from the binaries**: `collect_filter_columns` =
A 0 / B 1 / C 0 / D 1, `MaterializedHotStateBatch>::filter` = A 18 / B 18 / C 19 / D 19, four
distinct md5s. 5 reps, fresh process and fresh RocksDB dir each, arm order rotated every rep.

| cell | A base | B +#1453 | C +#1454 | D union |
|---|---|---|---|---|
| `range` N=1  **[both live]** | 140 936 µs | 125 391 (**−11.0%**) | 121 851 (**−13.5%**) | 106 067 (**−24.7%**) |
| `range` N=100 *[null for #1453]* | 191 367 | 194 268 (+1.5%) | 151 226 (−21.0%) | 152 393 (−20.4%) |
| `range` N=10 000 *[null for #1453]* | 193 937 | 195 962 (+1.0%) | 153 968 (−20.6%) | 154 027 (−20.6%) |
| `point` N=1 *[null for #1453]* | 27.9 | 28.7 (+2.9%) | 28.1 (+0.7%) | 27.9 (+0.1%) |

On `range` N=1 **every pair is disjoint**, including **B vs D and C vs D** — the union is provably
better than either change alone, not merely better on medians.

```
range N=1 raw per-rep µs
  A base    139218 140292 140936 141362 142378
  B +#1453  124998 125145 125391 125621 125729
  C +#1454  120975 121523 121851 123316 123379
  D union   105623 105845 106067 106302 106667
```

**Additivity:** B saves 15 545 µs, C saves 19 084 µs, sum 34 629 µs, **union actually saves
34 869 µs — 100.7% of the sum.** Clean additivity, not super-additivity: the two changes remove
disjoint work (one removes the DOM, the other removes the buffer clones) and neither unlocks extra
headroom in the other.

Two coherence checks that had to pass and did: on the null cells **D vs C is overlapping** (#1453 is
inert there, so the union must equal #1454 alone), and #1454 alone reproduces its own PR's ~−21% on
a fixture it was not tuned on.

**Answers byte-identical across all four arms and all reps** — one distinct `(fold, row_count)` per
cell, where the fold hashes every cell of every returned row.

### The recomposed profile — and an honest correction to my own headline

Self-time, gated read subtree, `range` N=1, T=200k:

| bucket | A base | D union |
|---|---|---|
| `bytes::Bytes` refcounting | 32.8% | **21.1%** |
| serde_json library | 24.2% | **6.0%** |
| *plus* `parse_snapshot_filter_columns` (my streaming parser) | — | **7.2%** |
| RocksDB iterator | 15.0% | 16.5% |
| lix key/value decode | 6.5% | 13.3% |
| **Arrow / columnar build** | **0.03%** | **0.06%** |

**The correction:** my bucket predicate counted only `serde_json::*` symbols, so "6.0%" understates
the JSON axis. The streaming parser is a lix symbol and costs a further **7.2%** — so JSON
tokenisation is **24.2% → ~13.2%**, not → 6%. Parsing less per row did not make parsing free: every
row's JSON text is still walked to find one field.

**Two caveats before anyone picks the next target from this.** `[unknown]` is **24.6%** of the
union's gated stacks versus 9.2% of the base's — the frame-pointer chain resolves materially worse
in the union binary, so the residual attribution is provisional and wants its own pass. And the
`lix key/value decode` rise is more likely inlining/attribution shift than real growth.

What is solid: **Arrow is still 0.06%.** Removing two thirds of the per-row overhead did not turn
this into a vectorised scan — it is still a per-row pipeline that materialises Arrow at the end.
The largest resolved leaves in the union are `bytes::promotable_even_drop` 11.3%,
`decode_head_value` 7.8%, `parse_snapshot_filter_columns` 7.2%, `shallow_clone_vec` 6.3%,
`decode_hot_scan_row_key_in_scope` 5.3% — i.e. residual refcount traffic (already being worked),
row decode, and the irreducible cost of walking JSON text per row.

PR #1454, opened 2026-08-13
perf(hot-state): stop cloning shared row buffers on every scanned row

samuelstroschein:
## What this is

A frame-pointer profile of a 200 000-row range scan, self time gated through
`execute_read_statement_with_store`, attributed **36% of the scan to `bytes::Bytes`
refcounting** and **0.08% to Arrow**. That share is atomic refcount traffic, not
data movement: a contended increment per clone and a decrement per drop, per row,
per scan.

This removes the three call sites responsible. **No format change** — nothing
persisted changes shape, and the diff touches no write path or encoding symbol.

## Where the clones were

Every one of them cloned a buffer the caller was about to discard.

| site | what it did |
|---|---|
| `MaterializedHotStateBatch::filter` | rebuilt a **second columnar owner row by row** through `push_ref`, cloning each surviving row's `SharedStr`s and each `EntityPk` component, then dropped the original |
| `apply_entity_batch_filters` (entity provider) | the same rebuild again, one layer up |
| `HotScanIdentity::push_materialized` | consumes `self`, yet called `push_materialized_ref`, which **cloned the primary key out of a value dropped on the next line** |

## What it does now

`filter` **consumes** the batch and compacts the columns it already owns against one
shared keep mask, so surviving rows are *moved*. The string dictionary is carried over
rather than rebuilt — it holds one entry per *distinct* schema key, file id and branch
id, not one per row, so a dropped row strands no per-row allocation. A new
`push_materialized_interned` interns the identity strings from borrows while *moving*
the key.

No `unsafe`, no new lifetime relationships, no borrow escaping a buffer's owner.

## Evidence

Machine `ryzen-9950x-I`. RocksDB, **mimalloc** as `#[global_allocator]`, 200 000 rows,
one checkpoint, `range` shape, 3 interleaved reps with arm-order rotation. Both arms
built from `0b97f3c45`; the candidate binary was verified to contain
`push_materialized_interned` and the baseline binary verified not to.

Raw per-rep milliseconds:

| n | baseline | candidate | median delta |
|---|---|---|---|
| 1 000 | 184.4 183.8 185.3 | 145.0 145.0 144.7 | **−21.4%** |
| 10 000 | 185.8 185.4 187.8 | 146.6 146.7 147.9 | **−21.0%** |
| 100 000 | 203.3 202.6 204.6 | 162.8 163.8 163.3 | **−19.7%** |

**Ranges are disjoint at every size.** A **null control** of two consecutive baseline
runs read 184.9–190.8 ms — overlapping the baseline, nowhere near the candidate.

**Both arms return byte-identical answers**: the bench's result fold is
`328159715142113358` in every run of both arms, including the frame-pointer runs.

### Same instrument before and after

`perf record -F 2999 --call-graph fp`, frame pointers and `+sse4.2,+pclmulqdq` via env
`RUSTFLAGS` into a separate target dir, self time gated to the read subtree:

| symbol | base | cand |
|---|---|---|
| `shared_clone` | 15.25% | **1.73%** |
| `shared_drop` | 11.02% | **4.72%** |
| `shallow_clone_vec` | 4.07% | 4.70% |
| `promotable_even_drop` | 3.38% | 4.84% |
| `promotable_even_clone` | 2.64% | **0.03%** |
| **total refcount** | **36.35%** | **16.02%** |

Those are shares of a pie that itself shrank. Per query, refcount work falls **66%**
(0.375 → 0.128 Gcycles/query; gated cycles/query 1.033 → 0.801, −22.4%, matching the
wall clock).

All three targeted callers disappear from the attribution:
`filter → shared_clone` 11.08% → 0, `push_ref → {promotable_even_clone, shared_drop}`
5.80% → 0, `push_materialized → shared_clone` 2.96% → 0.

The two symbols that did **not** fall are the promotion and drop of each row's own
physical key inside `decode_hot_scan_row_key_in_scope`. This change does not address
them and their absolute cost is flat; they are the next target on this axis.

### A trap worth recording

The first candidate frame-pointer build was a **no-op**: `cargo build` with a shared
`CARGO_TARGET_DIR` across two worktrees at the same sha reported `Finished in 0.12s`
and left the *baseline* binary in place, so the "candidate" profile was the baseline
profile. It was caught because the candidate's wall clock matched the baseline's to
0.05%. Every arm here uses a **dedicated target dir**, and arm identity is asserted by
symbol presence in the binary, not by which directory it came from.

## Not in scope

The other 30% of the scan is `serde_json` re-parsing each row's snapshot on every scan.
That needs typed columns in the packed base and therefore a `REPOSITORY_PROTOCOL_VALUE`
break; it is being scoped separately.

### DIFF
--- a/packages/e2e/Cargo.toml
+++ b/packages/e2e/Cargo.toml
@@ -522,3 +522,8 @@ required-features = ["sdk-tests", "storage-benches", "slatedb"]
 name = "exppth2_txn_probe"
 path = "examples/exppth2_txn_probe.rs"
 required-features = ["storage-benches"]
+
+[[test]]
+name = "hot_scan_refcount_census"
+path = "tests/hot_scan_refcount_census.rs"
+required-features = ["storage-benches"]
--- a/packages/e2e/tests/hot_scan_refcount_census.rs
+++ b/packages/e2e/tests/hot_scan_refcount_census.rs
@@ -1,0 +1,146 @@
+//! Deterministic census of refcounted-buffer handle clones on the HOT scan
+//! read path.
+//!
+//! **Why a count and not a nanosecond figure.** A rate only travels within a
+//! host class — this fleet spans a 5.4x gap between its slowest and fastest
+//! machines, so "585 ns per scanned row" is a fact about one box. The number
+//! of times the scan duplicates a refcounted buffer handle is identical
+//! everywhere, so it is the half of a "we stopped cloning" claim that can be
+//! re-run anywhere and compared across agents and rounds.
+//!
+//! **Own process on purpose.** The counters in `lix::storage_bench` are
+//! process-global; an exact assertion on them inside the shared suite reads
+//! whatever every other concurrently running test happened to scan. A
+//! dedicated `[[test]]` target is its own process, which is what makes the
+//! exact per-row equalities below legitimate rather than flaky.
+//!
+//! The census counters sit on the clone expressions themselves, not on the
+//! functions containing them, so a site that stops cloning stops counting.
+
+use lix::storage_bench::take_hot_scan_refcount_census;
+use lix::{Value, open_lix};
+
+const ROWS: usize = 2_000;
+const SCHEMA_KEY: &str = "hot_scan_refcount_census_probe";
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn a_full_scan_clones_a_bounded_number_of_shared_handles_per_row() {
+    let lix = open_lix().await.expect("open in-memory lix");
+
+    let schema = serde_json::json!({
+        "x-lix-key": SCHEMA_KEY,
+        "x-lix-primary-key": ["/id"],
+        "type": "object",
+        "properties": {
+            "id": { "type": "string" },
+            "ordinal": { "type": "number" },
+            "lane": { "type": "string" }
+        },
+        "required": ["id", "ordinal", "lane"],
+        "additionalProperties": false
+    });
+    lix.execute(
+        "INSERT INTO lix_registered_schema (value) VALUES (lix_json($1))",
+        &[Value::Text(schema.to_string())],
+    )
+    .await
+    .expect("register census schema");
+
+    for chunk in (0..ROWS).collect::<Vec<_>>().chunks(500) {
+        let mut sql = format!("INSERT INTO {SCHEMA_KEY} (id, ordinal, lane) VALUES ");
+        let mut params = Vec::new();
+        for (offset, ordinal) in chunk.iter().enumerate() {
+            if offset > 0 {
+                sql.push(',');
+            }
+            let base = offset * 3;
+            sql.push_str(&format!("(${}, ${}, ${})", base + 1, base + 2, base + 3));
+            params.push(Value::Text(format!("/census/{ordinal:09}")));
+            params.push(Value::Integer(i64::try_from(*ordinal).expect("fits i64")));
+            params.push(Value::Text(format!("lane-{:02}", ordinal % 8)));
+        }
+        let result = lix.execute(&sql, &params).await.expect("seed census rows");
+        assert!(
+            result.rows_affected() > 0,
+            "a seed statement that affects no rows produces the same zero counters as a route that never ran"
+        );
+    }
+    lix.create_checkpoint()
+        .await
+        .expect("drive the census fixture to packed-base steady state");
+
+    // Discard everything the seed and the checkpoint decoded.
+    let _ = take_hot_scan_refcount_census();
+
+    // A predicate on a non-key column, so the entity provider parses each
+    // row's snapshot and filters it — the shape that used to rebuild the whole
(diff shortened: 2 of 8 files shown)


samuelstroschein:
## Gate

Scope re-derived from `0b97f3c45`'s own `ci.yml` (cargo job): clippy `--profile test
--workspace --all-targets --all-features -- -D warnings`; `cargo test --profile test
--workspace --exclude lix_e2e --all-features --no-fail-fast`; `cargo test --profile test
-p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace
--no-fail-fast`. That sha has no `--no-default-features` step and no
`wasm32-unknown-unknown` step in the cargo job.

```
git rev-parse HEAD                 794822ed963ad303b9905c841bbbb97e1a257255
git status --porcelain             (empty, before and after)
CI_SCOPE_CONFIRMED_AT              0b97f3c45a1b106262a0b410aaf0ea403160cc1b
CLIPPY_EXIT=0  TEST1_EXIT=0  TEST2_EXIT=0
EXECUTED_TARGETS   test1=38  test2=27
TEST1  3432 passed / 0 failed      (reference 3430; +2 is exactly the two tests added here)
TEST2   172 passed / 0 failed      (reference 172)
NEW_TEST_PRESENT=1  filtering_moves_surviving_rows_instead_of_cloning_their_buffers
NEW_TEST_PRESENT=1  filtering_stops_calling_the_predicate_once_the_limit_is_reached
```

`filtering_moves_surviving_rows_instead_of_cloning_their_buffers` asserts the filtered
batch reuses the **same `EntityPk` column allocation** as the pre-filter owner, which is
what a row-by-row rebuild could not do.

samuelstroschein:
## Second commit: the deterministic count

`944694200` adds a census of refcounted-buffer handle clones on the HOT scan read
path, because **a nanosecond figure only travels within a host class** and this fleet
spans a 5.4× gap. A clone count is identical on every machine.

Every counter sits on the **clone expression itself**, not on the function containing
it, so a site that stops cloning stops counting and a site that is merely renamed keeps
counting. `packages/e2e/tests/hot_scan_refcount_census.rs` is a **dedicated `[[test]]`
target** — these counters are process-global, and an exact assertion inside the shared
suite reads whatever else happened to be scanning at the time.

Same 2 000-row full-scan lane, in-memory storage, one checkpoint, before and after:

| bucket | before | after | |
|---|---|---|---|
| rows decoded | 2 058 | 2 058 | denominator |
| key handle clones | 2 058 (**1.000/row**) | 2 058 (**1.000/row**) | unchanged |
| value handle clones | 2 046 (0.994/row) | 2 046 (0.994/row) | unchanged |
| **batch-rebuild clones** | **10 320 (5.015/row)** | **88 (0.043/row)** | **−99.1%** |
| **total** | **7.009/row** | **2.037/row** | **−70.9%** |

This supersedes the "≥6 per scanned row" figure in the first commit's message, which was
*inferred from the code* and labelled as such. The measured number is **4.97 batch-rebuild
handle clones eliminated per scanned row**.

**Verified by inversion, not by assertion alone.** The identical test file applied to the
pre-fix tree fails with

```
materialized batches should be compacted in place, not rebuilt row by row;
got 10320 batch-rebuild handle clones for 2058 decoded rows (5.015/row, ceiling 0.100/row)
```

so a green result is evidence the fix is present, not evidence the census never fired.
The ceiling is 0.100/row: fifty times above the observed value and fifty times below the
pre-fix value, so fixture drift cannot trip it and the rebuild coming back cannot survive
it. It is a threshold rather than an equality because the counters are process-global and
the statement under test is not the only thing in the process that materializes a row.

The two **unchanged** buckets are the promotion and drop of each row's own key and value
buffers — `Bytes::copy_from_slice` per key and per value in the adapter's `next_page`,
each promoted to a shared allocation on its first clone. That is the next target on this
axis, and this census is now the instrument for it.

## Gate, re-run at `944694200`

```
git rev-parse HEAD                 9446942006af29326433f7ead36623ee9a7414fa
git status --porcelain             (empty, before and after)
CI_SCOPE_CONFIRMED_AT              0b97f3c45a1b106262a0b410aaf0ea403160cc1b
CLIPPY_EXIT=0  TEST1_EXIT=0  TEST2_EXIT=0
EXECUTED_TARGETS   test1=38  test2=28        (was 27; +1 is the census target)
TEST1  3432 passed / 0 failed                 (reference 3430, +2 from commit 1)
TEST2   173 passed / 0 failed                 (reference 172, +1 from the census)
NEW_TEST_PRESENT=1  a_full_scan_clones_a_bounded_number_of_shared_handles_per_row  (… ok)
NEW_TEST_PRESENT=1  filtering_moves_surviving_rows_instead_of_cloning_their_buffers
NEW_TEST_PRESENT=1  filtering_stops_calling_the_predicate_once_the_limit_is_reached
```

Every count moves by exactly the number of tests and targets added, and each added test is
named in the executed log.

PR #1452, opened 2026-08-13
probe(lix): hot-scan decode census + tombstone accumulation under churn

samuelstroschein:
## What this adds

Instruments only. No production behaviour changes.

1. **A decode census inside the hot row scan loops.** `HOT_SCAN_DECODED_ENTRIES`,
   `HOT_SCAN_MATCHED_ENTRIES`, `HOT_SCAN_TOMBSTONE_ENTRIES`, incremented in the
   **per-entry decode loop** of both arms — `hot_scan_entries`' wide fallback and
   `scan_hot_file_entries` — not at the layer that returns the answer. Gated
   `#[cfg(any(test, feature = "storage-benches"))]`.
   Unlike the existing `lix_binary_blob_ref` census next to it, these are
   ungated on schema key, so they observe any request.

2. **Phase 7 (`hot_row_tombstone_decode_census`)** — what fraction of a scan's
   decode work is tombstones.

3. **Phase 11 (`hot_row_tombstone_churn_cycles`)** — whether a churning workload
   accumulates tombstones without bound, with **checkpoint cadence as the
   experimental variable** and the compaction engagement counters read per round.

## What they measured

Lane: `packages/lix` lib tests, `Memory` storage adapter, release profile,
`ryzen-9950x-IV`, single test thread. Answer is exactly **one row** in every arm
of every table below.

### Tombstones are essentially all of the decode work (phase 7)

| arm | n | ROW_SPACE entries | tombstones in space | decoded | decoded tombstones | scan µs |
|---|---|---|---|---|---|---|
| fresh | 10 000 | 43 | 0 | 1 | 0 | 31 |
| update_churn | 10 000 | 44 | 0 | 2 | 0 | 34 |
| churn | 10 000 | 10 042 | 9 999 | 10 000 | 9 999 | **2 881** |
| churn_dropped | 10 000 | 43 | 0 | 1 | 0 | **32** |

`update_churn` performs 9 999 updates — the same order of write volume and
generation rotations as `churn` — and decodes **2** entries. `churn_dropped` is
the same fixture as `churn` with the tombstones physically removed and the
engine reopened cold. Together they attribute the 90x to tombstones and not to
per-commit plane rematerialisation.

**99.99 % of the decoded entries are tombstones**, and removing them returns the
scan exactly to the no-history floor.

### The accumulation is unbounded, and checkpoint cadence does not help (phase 11)

500 rows created and deleted per round, one permanent survivor, 8 rounds.

[...]

PR #1458, opened 2026-08-13
perf(lix): never publish a tombstone for an interval-local delete

samuelstroschein:
**Stacked on #1452** (`agent/e60-tombstone-decode-census`, instruments-only). Base this on that branch, not on `claude-7-optimizations` directly. #1452 stays instruments-only; this PR is the behaviour change.

## The defect

An identity **created and deleted inside one checkpoint interval** is net-absent against the interval baseline. Its `ROW_SPACE` tombstone shadows nothing — and nothing ever reclaimed it.

Measured on #1452's instruments: the checkpoint's compaction route **runs and is offered nothing** for exactly this workload.

```
checkpoint_each_round,2,...,tombstones=1000,...,routes=1 offered=0 candidates=0 compacted=0
```

`hot_generation_has_any_base` was not the blocker — `packed_bases=0 root_bases=0` throughout, so gate (a) passes. The blocker is upstream of every gate: these identities never enter the dirty set a checkpoint republishes, so the mask never sees them. Tombstones accumulated **1:1 with deletes, forever**, and a collection scan decoded every one of them.

## The change

**This is not "run `hot_compaction_mask` on every commit."** That would trip gate (c) and fail the publication, because an ordinary commit's rows are not `(Clean, false)`. It is a **second, narrower admission rule at the same site**, sharing gates (a) and (b) unchanged, admitting only a delete whose pre-image already carries `BeforeAbsent` for the active checkpoint.

Two things make it cheap:

- **The discriminator is already computed.** `next_hot_working_diff_baseline` returns `(BeforeAbsent { checkpoint_commit_id: active }, newly_dirty = false)` for exactly the interval-local case. The precondition is read from the same pre-image the loop already holds — **no new store read, no new pass** — and the value loop reconciles the cheap predicate against the baseline it actually derived before dropping any row (gate (c) for this route).
- **Physical removal is established, not new.** `next_value_ranges.push(None)` is the path untracked deletes (`physically_deletes()`) already take in this plane.

## Why the tombstone is owed to nobody

The reader analysis is in #1452 (phase 12), removing exactly these tombstones behind the engine's back and reopening cold. The decisive measurement:

> `arm=local before: entries=84 tombstones=40 working_diff_rows=0 live=1`

**`lix_working_diff` already reports nothing for these identities while their tombstones exist.** Collection and point answers, the closing checkpoint, the commit graph, a branch forked mid-interval (41 live either way, served from its own scope over a root base), a merge of that fork back, and undo are all unchanged — each with a `remove=false` null control returning the same verdict.

## Measured

One-row answer throughout. `Memory` adapter, release profile, `ryzen-9950x-IV`. 500 rows created and deleted per round, 8 rounds:

| cadence | tombstones r8 before → after | scan µs r8 before → after |
|---|---|---|
| no checkpoint | 4 000 → **0** | 1 394 → **32** |
| checkpoint each round | 4 000 → **0** | 1 392 → **33** |
| checkpoint every 4 | 4 000 → **0** | 1 379 → **32** |
| delete deferred past a checkpoint | 0 → 0 | 32 → 32 (compaction's job, unchanged) |

Decode census at 10 000 create/delete, one-row answer:

| arm | decoded | of which tombstones | scan µs |
|---|---|---|---|
| fresh | 1 → 1 | 0 | 31 → 33 |
| **update_churn** (null control: same write volume, no deletes) | **2 → 2** | 0 | **34 → 35** |
| churn | 10 000 → **1** | 9 999 → **0** | 2 881 → **31** |
| churn_clean (Clean pre-image, elision refuses) | 10 000 | 9 999 | 3 264 |

The **null control is unchanged**, so this is not a plane-rematerialisation effect. `churn_clean` is the remaining exposure and it is the compaction route's job: a delete of a checkpointed row keeps its tombstone until the next checkpoint discharges it, which the `deferred_delete_ckpt` row above shows working.

## Both inversions asserted, non-`#[ignore]`d

- `clean_pre_image_delete_still_publishes_a_tombstone` — pre-image predates the interval (`Clean`). Tombstones survive (200) and the working diff reports all 200.
- `interval_local_delete_over_a_base_still_publishes_a_tombstone` — a branch forked from a published head serves over a root base (`root_bases=1`). The deltas **reach** the route (`candidates >= N`, so the refusal is not vacuous) and gate (a) refuses every one; all 50 tombstones survive.

Both refusals are asserted from the **`ROW_SPACE` footprint**, which is fixture-local, never from the elision counters, which are process-global — see the last commit: `assert_eq!(elided, 0)` read **251** for a fixture that offered 50 once the parallel suite ran alongside it. Only the `>=` direction of a global counter is bleed-safe, and `>=` cannot express a refusal.

## Contention warning

This touches `hot.rs`, which is heavily contended this round. If another PR lands in `hot_compaction_mask` or the value-staging loop, **gate the merge of both before either lands** — the signature of `hot_compaction_mask` changed (two new parameters).

## Gate

See the PR comment.


samuelstroschein:
## Gate — green, full five steps

```
git rev-parse HEAD                     7d3843d4a7cec6dd27f0d4f758112038e2112e22
git status --porcelain                 (empty)
CI_SCOPE_CONFIRMED_AT=7d3843d4a        re-derived from this sha's ci.yml
EXECUTED_TARGETS test1+test2=65        3605 passed, 0 failed
NEW_TEST_PRESENT=1                     all three acceptance tests appear in the executed log
```

```
cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings      0
cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast    0
cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast   0
cargo check -p lix --no-default-features                                                 0
cargo check -p lix_js_sdk --target wasm32-unknown-unknown                                0
```

The `lix_e2e` feature list was re-derived from this sha's `ci.yml` rather than carried
forward; it is `sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace`.

**Provenance note, stated plainly:** the gate ran on the working tree that this commit
records. No file changed between the gate finishing and the commit — `git status` was
empty immediately after `git commit`, and the commit introduced no edits of its own.

### The gate caught a real defect in my own tests

The previous run was **red** (`TEST1=101`, 2 failed), and it was the process-global
counter trap:

```
phase13 | inversion=has_base entries=144 tombstones=50 packed=0 root=1 routes=4 offered=301 candidates=301 elided=251
```

The fixture offered **50**. The counters read **301 offered / 251 elided** because the
parallel suite's other publications land in the same process-global statics. Both
inversions passed in isolation and failed in the suite. Fixed by asserting the refusals
from the fixture-local `ROW_SPACE` footprint and keeping counters only in the
bleed-safe `>=` direction. Worth recording because both tests were *correct about the
engine* and wrong about the instrument.

### Prior-tree note

An earlier five-step gate on `626d9c21c` (instruments only, no behaviour change) reported
65 targets / 3602 passed. This tree reports 65 targets / 3605 passed: **+3, exactly the
three acceptance tests added here**, and no target count moved. Every added test is
accounted for.

Not merged, per standing instruction.

PR #1461, opened 2026-08-13
probe(lix): measure what a JSON-column equality costs the entity provider

samuelstroschein:
Instruments only. No production behaviour change. Independent of #1452/#1458.

## What this adds

1. **`json_predicate_pushdown_probe.rs`** — three spellings of a one-row-answer equality on the same 10 000-row fixture, measured with `provider_rows_examined`.
2. **A census on `filterable_column_name`**, at the refusal itself, logging *schema key and column* rather than counting — the shape is the question, and a counter cannot say which schema. Inert unless `LIX_FILTERABLE_CENSUS` names a file, env lookup cached in a `OnceLock`.

## What they measured

`packages/lix` lib tests, `Memory` adapter, release profile, `ryzen-9950x-IV`, single test thread. Hot-plane lane, no checkpoint and no packed base. Answer is exactly **one row** in every arm.

| arm | answer rows | `provider_rows_examined` | `scan_rows` |
|---|---|---|---|
| **string** — `k = 'k-7'`, pushable | 1 | **10 000** | **1** |
| **json** — `v = lix_json('{"n":7}')`, refused | 1 | **10 000** | **10 000** |
| **primary key** — `id = 'r-7'` | 1 | **1** | 1 |

### The pushable arm is the point

A JSON-column equality cannot become an `EntityRowFilter`, because `filterable_column_name` returns `None` for `EntityColumnType::Json`. The consequence is visible in `scan_rows`: the provider emits all 10 000 rows and DataFusion's `FilterExec` discards 9 999 above it.

But **`provider_rows_examined` is 10 000 in the pushable arm too.** The stored-row examination is identical, so "all rows are materialised and discarded" is true and is **not** a consequence of the JSON refusal. `EntityRowFilter` at this layer is a **post-materialisation filter**: it runs after the rows are read and only saves Arrow construction plus a `FilterExec` pass for the rows it rejects.

`sql_profile.rs` states the contract this rests on, in the counter's own documentation:

> *"a fix that only moves filtering earlier in the plan changes `scan_rows` and leaves this unchanged"*

A JSON pushdown is exactly such a fix. **Predicted, not measured: `scan_rows` 10 000 → 1, `provider_rows_examined` unchanged at 10 000.** Anyone building it should hold it to that prediction.

### The primary-key arm is the reference point

`examined = 1` is what an access path looks like — four orders of magnitude from either equality arm. **The 10 000× examination cost is shared by every non-PK, non-indexed equality, JSON or not.** That is an access-path problem, not a pushdown problem.

A JSON column can never reach that access path: `derive_indexed_columns` (`sql2/catalog/entity_surface.rs`) admits only `String | Integer`, one layer away from the pushdown. So the large win is structurally unavailable to JSON columns regardless of what `filterable_column_name` returns.

## Why `provider_rows_examined` and not a local census

My first attempt counted at `apply_entity_batch_filters`, above its `filters.is_empty()` early return. **The JSON arm produced no line at all** — that function never ran for it. Entity surfaces take several routes, and `provider_rows_examined` is recorded at **six** sites for that reason, including the columnar and overlay routes.

So a "zero frames in `apply_entity_batch_filters`" reading is consistent with **different route**, not only with "the filter did nothing". That census is deliberately **not** in this PR; the probe uses the instrument that observes every route. It is also task-local rather than process-global, so it cannot bleed across a parallel suite.

## Frequency, with the caveat that matters

Census over both suites:

| suite | accept | refuse_json |
|---|---|---|
| workspace (38 targets, 3 430 tests) | 543 | 6 — all one synthetic test schema |
| `lix_e2e` (27 targets, 172 tests) | 414 | **2 — both `lix_key_value.value`** |

`lix_key_value.value` is a real JSON column, so the shape does occur outside a test schema. **But refusal count is the wrong metric** — one refusal on a 10 000-row table dominates hundreds on small ones — and this is corpus frequency, not production frequency. Neither number should be read as evidence of rarity.

## Not included

No pushdown. The soundness analysis says one is possible — the residual materialises through `entity_json_text_value` → `json_to_string`, so a provider-side push computing the same function over the same `JsonValue` is **bit-identical by construction rather than a superset** — but the measurement above says the win is bounded to Arrow construction plus a `FilterExec` pass, and it is unmeasured in time. Banking the instrument instead.

### DIFF
--- a/packages/lix/src/json_predicate_pushdown_probe.rs
+++ b/packages/lix/src/json_predicate_pushdown_probe.rs
@@ -1,0 +1,188 @@
+//! Does a JSON-column equality reach the entity provider?
+//!
+//! Not a product module. `filterable_column_name` returns `None` for
+//! `EntityColumnType::Json`, so a JSON-column equality cannot become an
+//! `EntityRowFilter`. What that costs is the open question: the predicate is
+//! still *answered*, by DataFusion's `FilterExec` one layer above the
+//! provider, so the rows are materialised and then discarded.
+//!
+//! The text spelling (`WHERE v = '{"n":1}'`) is a type error
+//! (`LIX_ERROR_TYPE_MISMATCH`) and never runs, so it cannot be the shape under
+//! test. This probe uses the spelling that type-checks,
+//! `WHERE v = lix_json('...')`, and a String-column arm as the control: same
+//! fixture size, same one-row answer, on a column that *is* pushable.
+//!
+//! Counted at the materialisation boundary (`apply_entity_batch_filters`),
+//! above its `filters.is_empty()` early return, because a profile of that
+//! function reads identically for "nothing to filter" and "never ran".
+
+use serde_json::json;
+
+use crate::engine::Engine;
+use crate::session::SessionContext;
+use crate::storage_adapter::Memory;
+
+async fn open_session() -> SessionContext<Memory> {
+    let storage = Memory::new();
+    Engine::initialize(storage.clone())
+        .await
+        .expect("engine should initialize");
+    let engine = Engine::new(storage.clone())
+        .await
+        .expect("engine should open");
+    engine.open_session().await.expect("session should open")
+}
+
+async fn register(session: &SessionContext<Memory>, key: &str) {
+    let schema = json!({
+        "x-lix-key": key,
+        "x-lix-primary-key": ["/id"],
+        "type": "object",
+        "properties": {
+            "id": { "type": "string" },
+            "k": { "type": "string" },
+            "v": { "type": "object" }
+        },
+        "required": ["id", "k", "v"],
+        "additionalProperties": false
+    });
+    session
+        .execute(
+            "INSERT INTO lix_registered_schema (value) VALUES (lix_json($1))",
+            &[crate::Value::Text(schema.to_string())],
+        )
+        .await
+        .expect("schema should register");
+}
+
+async fn seed(session: &SessionContext<Memory>, table: &str, count: usize) {
+    const CHUNK: usize = 250;
+    let mut index = 0;
+    while index < count {
+        let end = (index + CHUNK).min(count);
+        let values = (index..end)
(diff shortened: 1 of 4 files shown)


samuelstroschein:
## Gate — green, full five steps

```
git rev-parse HEAD                     10727c4eddc71a464be806e89726cbcc658d7b09
git status --porcelain                 (empty)
CI_SCOPE_CONFIRMED_AT=10727c4ed        re-derived from this sha's ci.yml
EXECUTED_TARGETS test1+test2=65        3602 passed, 0 failed
NEW_TEST_PRESENT=1                     the probe appears in the executed log
```

```
cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings      0
cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast    0
cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast   0
cargo check -p lix --no-default-features                                                 0
cargo check -p lix_js_sdk --target wasm32-unknown-unknown                                0
```

The probe is `#[ignore]`d, so the gate lists it without executing it; its output is in the PR body and was produced on this same tree.

### The first run was red, and the guard that caught it was the right one

```
module_layers::every_top_level_module_is_accounted_for
top-level modules are neither in MODULE_LAYERS nor in UNLAYERED_MODULES: json_predicate_pushdown_probe
```

A new top-level module must declare where it sits. Registered in `UNLAYERED_MODULES` as `("json_predicate_pushdown_probe", "cfg(test) measurement probe")`, matching its two sibling probes, and placed in the list's alphabetical position. Worth noting for anyone adding a probe module: **the layering guard, not the compiler, is what tells you.**

### Target-count accounting

65 targets and 3 602 passed, against 65 / 3 602 on the tree this branches from. **No target moved and no count moved**, which is expected: the probe is `#[ignore]`d and the census is a `cfg`-gated function with no test of its own. Every number is accounted for.

Instruments only. Do not merge.

PR #1465, opened 2026-08-13
Typed column slots for scalar-declared entity columns

samuelstroschein:
Typed column slots for scalar-declared entity columns: the stored-record format, its decode path, and the equality witness it owes. **The writer is not in this PR** — see *What is not here* before reading this as a shipped optimisation.

## What this is for

A three-arm harness (full parse / shipped column-selective parse / no parse) put the residue typed slots would capture at **255–320 ns/row, ~48–53% of total parse cost**, 7–10× that harness's artifact band.

The residue is **tokenization, not materialization**. The shipped column-selective fix removes 5 of 6 column materializations per row and **0 of ~177 bytes** of tokenizing — `parse_snapshot_filter_columns` still runs a deserializer across the whole document to find the columns the predicate names. The prize is on **all scanned rows**, not just rejected ones: since `049392ead` deleted `parsed_snapshots`, a surviving row's filter parse is thrown away too, and the zero-rejection lane still shows 13.6–14.4% disjoint.

## The format

`packages/lix/src/hot_state/typed_slots.rs`. A positional record: a 3-byte header, a 9-byte directory entry per declared column in schema order, then a payload area.

```text
 0        format version (1)
 1..3     declared column count (big endian u16)
 3..3+9n  directory, 9 bytes per declared column: tag, u32 offset, u32 length
 3+9n..   payload area
```

Column names are **not stored**. A reader resolves a name to a slot index once per batch against the schema layout, then indexes the directory. Reading a column costs a bounds check and a slice, and **no byte of the record is tokenized** — including the bytes of the columns the reader did not ask for.

This is only sound because **schemas are closed, and it is a mechanism rather than a convention**: `additionalProperties: false` is required-present-and-required-false in `packages/lix/src/schema/definition.json`, and `assert_entity_properties_have_projectable_types` rejects any property whose declared type does not resolve. 0 of 101 columns are untyped.

### Scalar columns only. No tagged any-slot.

86 of 92 stored columns are scalar and get typed payloads. The 6 JSON-declared columns keep their JSON text verbatim in a `Json` slot. `lix_key_value.value` and `json_pointer.value` are **arbitrary JSON by design** — think of it as a SQL `::json` column, not a value of unknown type — so the line is drawn at the declaration: scalar-declared gets a typed slot, JSON-declared stays JSON. A `Json` slot is a byte range a reader hands back or re-parses exactly as it does today. Nothing about it gets worse and nothing about it pretends to be typed.

`only_a_json_declared_column_reaches_a_parser_and_only_when_wanted` pins that the exemption is scoped to the column, not to the record: a predicate over the five scalar columns of a mixed schema still tokenizes nothing.

### Absent keys become null slots (the null fix, folded in)

A slot per declared column cannot express "absent", so encoding materializes **every** declared column and writes a null slot where the key was missing. Absent keys really do occur after normalization — `would_apply` (`catalog/snapshot.rs`) only visits properties that *have* defaults, so a nullable, default-less property that was never written stays missing from the map.

This is safe because the entity surface already erases the distinction: `entity_json_text_value` (`sql2/providers/entity.rs`) maps both `None` and `Some(JsonValue::Null)` to SQL `NULL`. And it does not disturb the write path's deliberate absent-vs-null distinction, because that distinction is consumed during normalization — defaults apply to absent keys and not to explicit nulls, pinned by `normalization_does_not_overwrite_explicit_null_with_default` — which runs strictly before a row is encoded here.

The record keeps `Absent` and `Null` as **distinct tags** anyway, so the write path's distinction stays readable out of the record rather than being destroyed by it. `absent_and_explicit_null_are_distinct_in_the_record_and_equal_at_the_surface` asserts both halves. Folding this in now rather than later is deliberate: doing it separately would break format twice.

### The equality witness

`working_diff_slot_fingerprint` (`hot_state/tracked_head.rs`) digests JSON text and `payload_equality` compares those 32-byte hashes to decide Equal/Different, so a typed layout owes a replacement per-row content fingerprint that agrees across the dirty/clean transition. `typed_slots_fingerprint` is blake3 over the record bytes, which is well-defined as a content fingerprint because the encoder is order-free: slots are written in layout order, there is no padding, and no map-iteration order enters the encoding anywhere. `fingerprint_agrees_on_equal_content_and_separates_unequal_content` encodes the same content from two different source key orders and asserts the **bytes** are equal, not just the hashes.

## Evidence

Deterministic counts, not timings. The brief's rule — establish the artifact band on your own harness before quoting any delta — is why there is no ns figure here: no harness was built, so no delta is claimed.

`reading_predicate_columns_from_a_typed_record_tokenizes_nothing` projects `{id, count}` out of a 6-column all-scalar record and asserts:

| | shipped column-selective parse | typed record |
|---|---|---|
| bytes tokenized per scanned row | **151** (the whole document) | **0** |
| JSON parser invocations | 1 | **0** |

The zero is a **mechanism, not an inspection**: every point on the decode path that hands bytes to a tokenizer increments a `cfg(test)` counter, and the test asserts on the counter. Both arms are asserted to produce the same values for the projected columns, so the count is not a win over a route that answered differently.

`canonical_json_reconstruction_is_byte_identical_to_the_normalized_text` is the property a staged landing depends on and the reason this PR exists in this shape: a typed record can serve a text consumer **byte-identically**, over explicit nulls, absent columns, escaped strings, non-UTF-8-trivial content, i64/u64/f64 boundaries and `1e10`-style exponents. Number *kind as written* is preserved rather than coerced to the declared type — an `Integer` column holding `1.0` renders back as `1.0` — because coercing would make reconstruction differ from the bytes it replaced.

`permuted_layout_changes_the_record` guards the positional addressing, and every round-trip fixture gives each field a **distinct** value: a round trip with colliding values cannot detect a permuted slot order.

## What is not here, and why

**The writer.** Nothing in this PR encodes a typed record into a stored row, so no read path is served by one yet and **no performance claim is made**. The cut was made here deliberately rather than landing a half-migrated writer.

The blocker is structural and worth recording. The snapshot payload is `str`-typed end to end — `HeadSlotView::Inline(&str)`, `MaterializedHotStateRow::snapshot_content: Option<SharedStr>` — while a typed record's fixed-width number payloads are not valid UTF-8. So typed slots need a **new byte-valued channel** through the head value and the materialized row; they cannot be smuggled through the existing text one. Wiring a decode branch keyed on the existing `&str` channel would have produced a branch that can never be taken, which is worse than an honest gap.

That plumbing is also where the blast radius lives: `snapshot_content` has **817 references across 40+ files**, so a hard cutover to typed-only storage is not a single change. Two follow-ups, in order:

1. **A byte channel plus the writer**, with the JSON text still reconstructed on demand for the 817 text consumers. `canonical_json_reconstruction_is_byte_identical_to_the_normalized_text` is what makes this possible at all — content addressing hashes the JSON text, so a reconstruction that is not byte-identical is a correctness break, not a perf regression.
2. **Migrate the text consumers**, then drop the reconstruction.

Also not here: `parse_snapshot_filter_columns` is untouched, `HEAD_VALUE_VERSION` is untouched, and `storage_codec.rs` and the commit-mutation-directory record are untouched (owned elsewhere this round).

## Protocol

**`REPOSITORY_PROTOCOL_VALUE` is left at `tracked-default-branch.v68` on purpose.** This change is inert on stored bytes today, but the follow-up that lands the writer **requires the bump**, and all of this round's format changes are meant to land under a single bump applied once at integration. Four separate bumps would be four mutually incompatible repositories.

## Gate

Scope re-derived from `898879893`'s own `ci.yml`, then run wider: `cargo check -p lix --no-default-features` is **not** a CI step at that sha and neither is a `--test-threads=1` test lane. Gate evidence is posted as a comment on this PR.

### DIFF
--- a/packages/lix/src/hot_state/mod.rs
+++ b/packages/lix/src/hot_state/mod.rs
@@ -4,6 +4,7 @@ mod entity_columnar_cache;
 mod entity_decoded_column_cache;
 mod reader;
 mod tracked_head;
+pub(crate) mod typed_slots;
 #[cfg(test)]
 pub(crate) use tracked_head::{head_decode_entity_pk_probe, hot_decode_entity_pk_probe};
 mod types;
--- a/packages/lix/src/hot_state/typed_slots.rs
+++ b/packages/lix/src/hot_state/typed_slots.rs
@@ -1,0 +1,983 @@
+//! Typed column slots for scalar-declared entity columns.
+//!
+//! Current-state snapshots are stored as UTF-8 JSON text. Every read that needs
+//! one named column still runs a deserializer across the whole document: the
+//! column-selective filter parse removes five of six *materializations* per row
+//! and zero of the ~177 tokenized bytes. This module replaces the tokenizing
+//! step with an O(1) indexed lookup.
+//!
+//! Schemas in this repository are closed — `additionalProperties: false` is
+//! required-present-and-required-false in `schema/definition.json`, and
+//! `assert_entity_properties_have_projectable_types` rejects any property whose
+//! declared type does not resolve. So the set of columns a snapshot may carry is
+//! known from the schema alone, and a record can address them *positionally*
+//! instead of by key.
+//!
+//! ```text
+//!  0        format version (1)
+//!  1..3     declared column count (big endian u16)
+//!  3..3+9n  directory, 9 bytes per declared column, in layout order:
+//!               [0]    slot tag
+//!               [1..5] payload offset within the payload area (big endian u32)
+//!               [5..9] payload byte length (big endian u32)
+//!  3+9n..   payload area
+//! ```
+//!
+//! Column *names* are not stored. A reader resolves a name to a slot index once
+//! per batch against the schema layout, then indexes the directory directly, so
+//! the per-row cost of reading a column is a bounds check and a slice — no
+//! bytes of the record are tokenized, including the bytes of the columns the
+//! reader did not ask for.
+//!
+//! # Scalar columns only, by design
+//!
+//! `String`, `Integer`, `Number` and `Boolean` columns get typed payloads.
+//! `Json`-declared columns keep their JSON text verbatim in a `Json` slot:
+//! `lix_key_value.value` and `json_pointer.value` are arbitrary JSON by design
+//! — a `::json` column, not a value of unknown type — and this format
+//! deliberately does not introduce a tagged any-slot to cover them. A `Json`
+//! slot is a byte range a reader can hand back or re-parse exactly as it does
+//! today; nothing about it gets worse, and nothing about it pretends to be
+//! typed.
+//!
+//! # Absent keys become null slots
+//!
+//! A slot per declared column cannot express "absent". Absent keys really do
+//! occur after normalization — `would_apply` only visits properties that have
+//! defaults, so a nullable, default-less property that was never written stays
+//! missing from the snapshot map. Encoding therefore materializes *every*
+//! declared column and writes a null slot where the key was absent.
+//!
+//! This is safe at the entity surface because that surface already erases the
+//! distinction: `entity_json_text_value` maps both `None` and
+//! `Some(JsonValue::Null)` to SQL `NULL`. It does not disturb the write path's
+//! deliberate absent-vs-null distinction either, because that distinction is
+//! consumed during normalization (defaults apply to absent keys and not to
+//! explicit nulls, pinned by
+//! `normalization_does_not_overwrite_explicit_null_with_default`) which runs
+//! strictly before a row is encoded here. `absent_columns` is retained so the
+//! difference remains *readable* out of the record rather than being destroyed
+//! by it.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use serde_json::{Map as JsonMap, Number as JsonNumber, Value as JsonValue};
+
+/// Wire format version for a typed-slot record.
+///
+/// This is a physical layout constant, not a repository protocol value. A
+/// change to it requires a `REPOSITORY_PROTOCOL_VALUE` bump like any other
+/// stored-format change.
+pub(crate) const TYPED_SLOTS_VERSION: u8 = 1;
+
+const HEADER_BYTES: usize = 3;
+const DIRECTORY_ENTRY_BYTES: usize = 9;
+
+const TAG_NULL: u8 = 0;
+const TAG_ABSENT: u8 = 1;
+const TAG_FALSE: u8 = 2;
+const TAG_TRUE: u8 = 3;
+const TAG_I64: u8 = 4;
+const TAG_U64: u8 = 5;
+const TAG_F64: u8 = 6;
+const TAG_STR: u8 = 7;
+const TAG_JSON: u8 = 8;
+
+
+// Counted at every point on the decode path that hands bytes to a JSON
+// tokenizer, so a test can assert that reading columns out of a typed record
+// tokenized nothing rather than inspecting the code and believing it. Only a
+// `Json`-declared slot can move this counter; a scalar slot decodes to a value
+// without a parser ever seeing its bytes.
+#[cfg(test)]
+thread_local! {
+    static TYPED_SLOTS_JSON_PARSES: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+}
+
+#[cfg(test)]
+fn record_typed_slots_json_parse() {
+    TYPED_SLOTS_JSON_PARSES.with(|count| count.set(count.get() + 1));
+}
+
+#[cfg(not(test))]
+fn record_typed_slots_json_parse() {}
+
+#[cfg(test)]
+fn reset_typed_slots_json_parse_count() {
+    TYPED_SLOTS_JSON_PARSES.with(|count| count.set(0));
+}
+
+#[cfg(test)]
+fn typed_slots_json_parse_count() -> usize {
+    TYPED_SLOTS_JSON_PARSES.with(std::cell::Cell::get)
+}
+
+/// The declared type of a column, as the schema catalog resolves it.
+///
+/// Deliberately a local enum rather than a re-export of the SQL surface's
+/// `EntityColumnType`: this module sits below the catalog, and importing the
+/// catalog here would invert that layering for no gain. The SQL side maps its
+/// own enum onto this one at the boundary.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum DeclaredType {
+    String,
+    Integer,
+    Number,
+    Boolean,
+    /// Arbitrary JSON by design. Stored as verbatim JSON text.
+    Json,
+}
+
+impl DeclaredType {
+    pub(crate) fn is_scalar(self) -> bool {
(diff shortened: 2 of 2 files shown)


samuelstroschein:
## Gate — green

Host `ryzen-9950x-I`, private worktree `~/claude5/tmp/ts-slots`, run nonce `3030720-1786646528`. The gate script asserts its own expected sha before running anything and prints `HEAD` at both ends.

```
HEAD_BEFORE=c798cd11b8450e85fc31bbdd361a96a091a63fae
DIRTY=0
CI_SCOPE_CONFIRMED_AT=c798cd11b8450e85fc31bbdd361a96a091a63fae
HEAD_AFTER=c798cd11b8450e85fc31bbdd361a96a091a63fae
```

Scope re-derived from `898879893`'s own `ci.yml`, then run **wider**:

| step | in CI at that sha | exit |
|---|---|---|
| `cargo check -p lix --no-default-features` | **no** | 0 |
| `cargo check -p lix --tests` | yes | 0 |
| `cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings` | yes | 0 |
| `cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast -- --test-threads=1` | yes, but **not** single-threaded | 0 |
| `cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast` | yes | 0 |

`--no-default-features` is the step that has caught this round's only real merge-only break and is not a CI step here. `--test-threads=1` is applied to the whole test1 lane, not just the new target: this change adds a `cfg(test)` thread-local parse counter, and a nondeterministic pass is not a pass.

### Counts reconcile exactly against a prediction stated before the run

Base at `898879893`, from that sha's own gate log: **executables 25 / passed 3478**, test2 **27 / 167**.

| | base | this branch | expected |
|---|---|---|---|
| test1 executables | 25 | **25** | 25 — no `[[test]]` target added |
| test1 passed | 3478 | **3491** | 3478 + 13 |
| test2 executables / passed | 27 / 167 | **27 / 167** | unchanged |

+13 is the 13 new tests, none of them `#[ignore]`d.

### `NEW_TEST_PRESENT=13`, asserted by emitted name

A non-zero target count proves the gate ran, not that these tests ran inside it. All 13 are grepped out of the executed log by emitted name:

```
hot_state::typed_slots::tests::absent_and_explicit_null_are_distinct_in_the_record_and_equal_at_the_surface
hot_state::typed_slots::tests::a_layout_with_a_duplicate_column_is_rejected
hot_state::typed_slots::tests::an_all_scalar_layout_is_reported_as_such
hot_state::typed_slots::tests::an_undeclared_predicate_column_projects_to_nothing
hot_state::typed_slots::tests::a_scalar_column_holding_a_composite_value_is_rejected
hot_state::typed_slots::tests::a_truncated_record_is_rejected_rather_than_read_out_of_bounds
hot_state::typed_slots::tests::a_typed_record_is_distinguishable_from_json_text_without_parsing_either
hot_state::typed_slots::tests::canonical_json_reconstruction_is_byte_identical_to_the_normalized_text
hot_state::typed_slots::tests::fingerprint_agrees_on_equal_content_and_separates_unequal_content
hot_state::typed_slots::tests::only_a_json_declared_column_reaches_a_parser_and_only_when_wanted
hot_state::typed_slots::tests::permuted_layout_changes_the_record
hot_state::typed_slots::tests::reading_predicate_columns_from_a_typed_record_tokenizes_nothing
hot_state::typed_slots::tests::typed_slots_round_trip_every_declared_column_with_distinct_values
```

### One check in the script is mis-specified — reporting it rather than quoting it

The script prints `FOREIGN_SHA_IN_LOG=1`. That is **not** a foreign-log detection: the grep looks for `ts-slots`, which is this job's own worktree path, so it matches its own log by construction and would report `1` on a clean run and on a clobbered one alike. It carries no information in either direction. Provenance here rests on the sha assertion, the run nonce in the log prefix, and `HEAD_BEFORE == HEAD_AFTER` instead. The check is fixed for the next run; it is recorded here because a summary line that can only ever print one value is a check that is not yet a check.

### Not gated

No timing number is produced, so no artifact band was established — deliberately. The evidence in this PR is deterministic counts (bytes tokenized per scanned row, parser invocations per row), which survive the 5.4× host gap that a ns/row figure would not.

samuelstroschein:
Merged to `main` after a wider-than-CI gate on the two-PR union (#1465 + #1467). #1468 (`e-tswriter`) was deliberately **excluded**.

**Union gated at `0ded298db`** (= `main`@`9e8497bc4` + #1465 + #1467, both merges conflict-free), **`main` re-gated at `f36b3d845`** from a fresh checkout. Trees are byte-identical (`git diff <union> origin/main` → 0 files). Host `ryzen-9950x-I`.

Scope re-derived from the merge target's own `ci.yml`, then run wider — six steps, all `EXIT=0` on both trees:

```
cargo check -p lix --no-default-features                                  EXIT=0
cargo check -p lix --tests                                                EXIT=0
cargo check --workspace --all-targets --all-features                      EXIT=0   (not a CI step)
cargo clippy --profile test --workspace --all-targets --all-features -D warnings   EXIT=0
cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast -- --test-threads=1   EXIT=0
cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast   EXIT=0
```

```
test1  25 executables / 3491 passed / 34 ignored / 0 failed
test2  27 executables /  167 passed / 21 ignored / 0 failed
```

**Counts reconciled by name, not by total** (`main`@`898879893` = 3478 → −1 #1464 → +13 #1465 → +1 passed / +2 ignored #1467 = **3491 / 34**, predicted before the run):

- #1465 — all **13** `hot_state::typed_slots::tests::*` present and `... ok`.
- #1467 — `dirnode_v2_summary_matches_walked_summary_and_rejects_touched_disorder ... ok`, plus `dirnode_point_read_cost_curve` and `dirnode_point_lookup_v1_vs_v2` `... ignored, measurement probe, not a gate` (+1 passed / +2 ignored).
- #1468 — its **five** test names appear **zero** times in the executed log. That absence, not the total, is the proof #1468 did not come along: the earlier four-way union read 3496 = 3491 + 5.

**#1468 exclusion asserted mechanically**, since `e-tswriter` was stacked on `e-tslots` and merging the wrong ref would have dragged it in silently:

```
git merge-base --is-ancestor 2bc45aaaf c798cd11b   -> exit 1 (FALSE)   # not on the e-tslots tip merged
git merge-base --is-ancestor 2bc45aaaf 3b7b85e95   -> exit 1 (FALSE)
git merge-base --is-ancestor 2bc45aaaf origin/main -> exit 1 (FALSE)   # after the merges
git merge-base --is-ancestor c798cd11b 2bc45aaaf   -> exit 0 (TRUE)    # confirms the stack direction
```

**No protocol bump.** `REPOSITORY_PROTOCOL_VALUE` reads `tracked-default-branch.v68` on `main` at `f36b3d845`, unchanged. Neither of these two PRs requires a bump; only #1468 does, and it is not merged.

Union diff vs `main` is purely additive — 3 files, 1867 insertions, 0 deletions.

PR #1467, opened 2026-08-13
measure: commit mutation directory node point-read amplification, and a seekable node design

samuelstroschein:
## What this is

A **measurement and a design**, not a shipped optimisation. Every line added here is `#[cfg(test)]`; no shipped code path changes and `REPOSITORY_PROTOCOL_VALUE` is untouched.

**The implementation this evidence argues for requires the round's protocol bump.** This PR does not, because a design is not a stored format — only shipped code is. That distinction is why it can land ahead of the bump.

## The finding

On a cold single-key point read, the record `commit mutation directory node` (`StoredNode`, decoded at `mutation_directory.rs::decode_node`) is materialised in full to serve one entry. It is a fanout-128 authenticated B-tree, so the *descent* is already seekable — only nodes on the path are fetched — but **decoding a node is all-or-nothing**: `Vec<StoredEntry>` / `Vec<StoredChild>` allocates an owned `Vec<u8>` per key bound for every entry in the node, and the read plan then walks all of them linearly to find the one it wants.

A codec swap cannot fix this. The shape is wrong.

## Instrument

A **task-local** census at `decode_node` — the process-global `read_accounting` counters in this module carry a comment saying they are unsuitable for assertions in a parallel suite, which is correct — plus a thread-local pass-through counting global allocator, so allocations are measured rather than derived. The allocator delegates to `System`, which is what this crate already used, since it declares none otherwise. Two reps per size, asserted equal, before any number was used.

## The growth curve — measured, ten sizes

Single-key point read, one directory root, `ryzen-9950x-II`:

| entries | ≈ rows | height | decodes | bytes decoded | entries materialised | allocations |
|---|---|---|---|---|---|---|
| 2 | 1 k | 1 | 1 | 134 | 2 | 5 |
| 8 | 4 k | 1 | 1 | 518 | 8 | 17 |
| 32 | 16 k | 1 | 1 | 2 054 | 32 | 65 |
| 64 | 33 k | 1 | 1 | 4 102 | 64 | 129 |
| 128 | 66 k | 1 | 1 | 8 199 | 128 | 257 |
| 256 | 131 k | 2 | 2 | 8 397 | 130 | 262 |
| 512 | 262 k | 2 | 2 | 8 587 | 132 | 266 |
| 2 048 | 1.0 M | 2 | 2 | 9 727 | 144 | 290 |
| 8 192 | 4.2 M | 2 | 2 | 14 287 | 192 | 386 |
| 16 384 | 8.4 M | 2 | 2 | 20 368 | 256 | 514 |

The row axis is `REPLACEMENT_PART_MAX_ROWS = 512` — one directory entry covers up to 512 rows.

### The shape is a sawtooth, not a straight line

Entries materialised per point read is `ceil(n / FANOUT) + leaf_size`: **linear in table size within each tree-height regime, dropping at each split, with the envelope growing as `FANOUT × height`.** A two-point measurement below 65 k rows sits entirely inside the first and steepest regime, which is why it reads as pure linearity. For the same reason "the decode count stays flat" is a regime artifact — decodes double at the first split.

What survives intact: **the amplification factor is 64–128× and does not decay with size.** At 8.4 M rows one point read decodes 20 KB and performs 514 heap allocations to return one entry.

## The design — "LXMD2", a seekable node

Fixed-stride tables over the node's own bytes, all little-endian:

```
  0  magic "LXMD2"        (5)
  5  kind                 (1)   0 = leaf, 1 = internal
  6  layout               (1)
  7  level                (2)
  9  count                (4)
 13  entry_count          (4)
 17  direct_row_count     (8)
 25  key_off[2*count + 1] (4 each)  alternating first_key/last_key bounds
     fixed[count]         (leaf 2 B; internal 44 B = node_id[32] + cumulative
                           entry_count + cumulative direct_row_count)
     key_region
```

A point read binary-searches the key table **in place** over `&[u8]` slices and decodes exactly one entry. Three consequences, only the first of which was the original goal:

1. Bytes read per node become `O(log fanout)` instead of `O(fanout)`, and allocations become constant.
2. Index routing — the path the keyless layouts (compact replacement, direct rows) take — becomes a binary search on the cumulative counts instead of a prefix-sum walk over every child.
3. `validate_loaded_node` becomes **O(1)**: the summary a parent authenticates is readable from the header and the two outermost key slots instead of being recomputed by walking the node.

## Measured — both node kinds, same allocation counter

Every arm asserts both forms return **the same answer at the same index** before reporting anything.

**Leaf node**, one point lookup:

| entries | v1 bytes read | v1 allocs | v1 alloc bytes | v2 bytes read | v2 allocs | v2 alloc bytes | node size |
|---|---|---|---|---|---|---|---|
| 2 | 134 | 5 | 380 | 167 | 2 | 54 | +18 B |
| 8 | 518 | 17 | 1 520 | 237 | 2 | 54 | +18 B |
| 32 | 2 054 | 65 | 6 080 | 307 | 2 | 54 | +18 B |
| 64 | 4 102 | 129 | 12 160 | 342 | 2 | 54 | +18 B |
| **128** | **8 199** | **257** | **24 320** | **377** | **2** | **54** | **+17 B (+0.2 %)** |

**Internal node**, one child lookup:

| entries | v1 bytes read | v1 allocs | v1 alloc bytes | v2 bytes read | v2 allocs | v2 alloc bytes | node size |
|---|---|---|---|---|---|---|---|
| 2 | 198 | 5 | 300 | 151 | 0 | 0 | +18.7 % |
| 8 | 768 | 17 | 1 200 | 221 | 0 | 0 | +13.5 % |
| 32 | 3 048 | 65 | 4 800 | 291 | 0 | 0 | +12.1 % |
| 64 | 6 088 | 129 | 9 600 | 326 | 0 | 0 | +11.8 % |
| **128** | **12 169** | **257** | **19 200** | **361** | **0** | **0** | **+11.7 %** |

At a full node: **21.7× fewer bytes on the leaf, 33.7× on the internal node; 128× fewer allocations on the leaf and zero on the internal node**, because a child summary is a fixed-stride record read onto the stack. v2 bytes grow `O(log n)` — one ~35 B key compare per doubling — and are flat in allocations.

**A full height-2 point read at 8.4 M rows is now measured on both halves rather than projected:** ~738 bytes and 2 allocations, against 20 368 bytes and 514 allocations today.

### Where v2 costs more, stated plainly

- **Internal nodes are ~11.7 % larger.** They carry both key bounds per child plus a 44-byte fixed record. Internal nodes are roughly 1/128 of the tree, so whole-tree growth is about **+0.3 %**. A standard separator-key-only internal node would remove most of this and was not tried.
- **The write side costs one extra allocation, flat**, and ~22 % more transient bytes during encode (33 788 vs 27 646 at 128 entries). Both come from an unoptimised two-pass encoder that builds a separate offset vector before writing; a single pass into a pre-sized buffer removes them. This mattered enough to measure rather than assume, because warm reads decode nothing — **the write path is the only place this layout can cost steady-state throughput.**

## Warm path

No regression is possible from this PR — it is entirely `#[cfg(test)]`. For the implementation: warm reads perform **zero** decodes, so they cannot regress from a decode change. The exposure is the write side, measured above.

## The authentication invariant

`validate_loaded_node` is what makes the tree authenticated: the summary a parent vouches for must equal the summary the loaded node actually has. v1 recomputes it by walking every entry; v2 reads it from the header.

`dirnode_v2_summary_matches_walked_summary_and_rejects_touched_disorder` is a **real gate, not a probe**. It asserts the header-read summary equals what `node_summary_ref` computes by walking, at 1/2/8/128 entries; that the read stays O(1) in bytes touched; and that a malformed key range inside the region the reader touches is refused.

What it does **not** cover, and the implementation must confront: lazy validation means a node's *unvisited* region is no longer structurally checked at read time. That is defensible — `load_nodes` verifies the content digest before decoding and the id comes from an authenticated parent, so an unvisited malformed region cannot affect the answer — but it is a real weakening of a read-time check into a build-time one, and it should be argued in the implementation's own PR rather than assumed here.

The fixture for the internal arm initially minted an all-zero `node_id` for child 0 and `validate_node` rejected it outright. A fixture bug rather than a finding — but the right kind, since the check that caught it is one the migration must preserve.

## What lands next, and its blast radius

The implementation is smaller than a rewrite of this file. `StoredNode` / `StoredEntry` / `StoredChild` stay as the builder's in-memory types; **only `encode_node` and `decode_node` change, plus a `NodeView` used by the selective read path.** The full-traversal paths (`load_all_mutation_part_read_plans`, `collect_mutation_directory_node_ids`, `decode_built_mutation_directory`) need every entry anyway and can keep materialising a `StoredNode`.

Every external caller — `gc.rs` (×3), `tracked_state/storage.rs` (×17), `tracked_state/mod.rs` — goes through function signatures and the `MutationDirectoryRoot` struct that the change leaves untouched, so **the blast radius stays `packages/lix/src/tracked_state/mutation_directory.rs` alone.**

## Reproducing

The two probes are `#[ignore = "measurement probe, not a gate"]` and contribute +0 to the passed count. The invariant test is not ignored and runs in the normal suite.

```
cargo test -p lix --lib tracked_state::mutation_directory::tests::dirnode_ \
  -- --ignored --nocapture --test-threads=1
```

### DIFF
--- a/packages/lix/src/tracked_state/mutation_directory.rs
+++ b/packages/lix/src/tracked_state/mutation_directory.rs
@@ -28,6 +28,288 @@ const NODE_MAGIC: &[u8] = b"LXMD1";
 const NODE_HASH_CONTEXT: &str = "lix commit mutation directory node v1";
 const ROOT_HASH_CONTEXT: &str = "lix commit mutation directory root v1";
 const FANOUT: usize = 128;
+
+// A pass-through counting allocator, test builds only.  It delegates every
+// call to `System` (which is what this crate already used, since it declares
+// no allocator otherwise) and counts only while a measurement scope is open
+// on the *current thread*, so a parallel sibling test cannot contaminate it.
+#[cfg(test)]
+mod alloc_census {
+    use std::alloc::{GlobalAlloc, Layout, System};
+    use std::cell::Cell;
+
+    thread_local! {
+        static ON: Cell<bool> = const { Cell::new(false) };
+        static ALLOCS: Cell<u64> = const { Cell::new(0) };
+        static BYTES: Cell<u64> = const { Cell::new(0) };
+    }
+
+    pub(super) struct Counting;
+
+    fn bump(size: usize) {
+        let _ = ON.try_with(|on| {
+            if on.get() {
+                let _ = ALLOCS.try_with(|counter| counter.set(counter.get().saturating_add(1)));
+                let _ = BYTES.try_with(|counter| {
+                    counter.set(counter.get().saturating_add(size as u64))
+                });
+            }
+        });
+    }
+
+    unsafe impl GlobalAlloc for Counting {
+        unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+            bump(layout.size());
+            unsafe { System.alloc(layout) }
+        }
+
+        unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+            bump(layout.size());
+            unsafe { System.alloc_zeroed(layout) }
+        }
+
+        unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
+            bump(new_size);
+            unsafe { System.realloc(ptr, layout, new_size) }
+        }
+
+        unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+            unsafe { System.dealloc(ptr, layout) }
+        }
+    }
+
+    /// Whether allocation counting is armed on this thread.  The decode census
+    /// consults this so its own bookkeeping never lands in a measured count.
+    pub(super) fn armed() -> bool {
+        ON.try_with(|on| on.get()).unwrap_or(false)
+    }
+
+    /// Runs `body` with allocation counting armed on this thread and returns
+    /// `(output, allocations, allocated_bytes)`.
+    pub(super) fn measure<T>(body: impl FnOnce() -> T) -> (T, u64, u64) {
+        let _ = ALLOCS.try_with(|counter| counter.set(0));
+        let _ = BYTES.try_with(|counter| counter.set(0));
+        let _ = ON.try_with(|on| on.set(true));
+        let output = body();
+        let _ = ON.try_with(|on| on.set(false));
+        let allocs = ALLOCS.try_with(|counter| counter.get()).unwrap_or(0);
+        let bytes = BYTES.try_with(|counter| counter.get()).unwrap_or(0);
+        (output, allocs, bytes)
+    }
+}
+
+#[cfg(test)]
+#[global_allocator]
+static COUNTING_ALLOCATOR: alloc_census::Counting = alloc_census::Counting;
+
+/// Prototype of the seekable node layout ("LXMD2").
+///
+/// A v1 node is one `musli(packed)` record: decoding it materializes every
+/// entry in the node, including an owned `Vec<u8>` per key bound, before the
+/// reader can find the single entry a point read wants.  A v2 node keeps the
+/// same information but stores it as fixed-stride tables over the node's own
+/// bytes, so a point read binary-searches the key table in place and decodes
+/// exactly one entry.
+///
+/// Layout, all little-endian:
+/// ```text
+///   0  magic "LXMD2"        (5)
+///   5  kind                 (1)   0 = leaf, 1 = internal
+///   6  layout               (1)
+///   7  level                (2)
+///   9  count                (4)   entries/children in this node
+///  13  entry_count          (4)   leaf entries under this node
+///  17  direct_row_count     (8)
+///  25  key_off[2*count + 1] (4 each)  alternating first_key/last_key bounds
+///      fixed[count]         (leaf: 2 bytes; internal: 44 bytes)
+///      key_region
+/// ```
+/// The node summary a parent authenticates (`first_key`, `last_key`,
+/// `entry_count`, `direct_row_count`) is readable from the header and the two
+/// outermost key slots, so validating a loaded node is O(1) rather than a walk
+/// over every entry.
+///
+/// The prototype covers `LAYOUT_BOUNDED_DIRECT` with no replacement part -
+/// the shape a cold point read actually traverses.  The other three layouts
+/// carry no key bytes at all and route by index, which the same cumulative
+/// tables serve.
+#[cfg(test)]
+pub(crate) mod seekable_prototype {
+    use super::*;
+
+    pub(crate) const V2_MAGIC: &[u8] = b"LXMD2";
+    const HEADER: usize = 25;
+    const LEAF_STRIDE: usize = 2; // direct_row_count u16
+    const INTERNAL_STRIDE: usize = 44; // node_id[32] + cum_entry_count u32 + cum_direct_rows u64
+
+    fn put_u32(out: &mut Vec<u8>, value: u32) {
+        out.extend_from_slice(&value.to_le_bytes());
+    }
+
+    fn header(
+        out: &mut Vec<u8>,
+        kind: u8,
+        layout: u8,
+        level: u16,
+        count: u32,
+        entry_count: u32,
+        direct_row_count: u64,
+    ) {
+        out.extend_from_slice(V2_MAGIC);
+        out.push(kind);
+        out.push(layout);
+        out.extend_from_slice(&level.to_le_bytes());
+        put_u32(out, count);
+        put_u32(out, entry_count);
+        out.extend_from_slice(&direct_row_count.to_le_bytes());
+    }
+
+    /// Encodes a bounded leaf in the v2 layout.
+    pub(crate) fn encode_leaf(layout: u8, entries: &[(Vec<u8>, Vec<u8>, u16)]) -> Vec<u8> {
+        let count = entries.len();
+        let direct_row_count: u64 = entries.iter().map(|entry| u64::from(entry.2)).sum();
+        let mut out = Vec::new();
+        header(
+            &mut out,
+            0,
+            layout,
+            0,
+            count as u32,
+            count as u32,
+            direct_row_count,
+        );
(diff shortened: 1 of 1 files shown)


samuelstroschein:
## Gate

`ryzen-9950x-II`, worktree `~/claude5/tmp/dirnode`, clean tree.

```
HEAD_BEFORE=3b7b85e95704ca2b1cd061a9885905bc62968fb5
HEAD_AFTER =3b7b85e95704ca2b1cd061a9885905bc62968fb5
DIRTY=0
CI_SCOPE_CONFIRMED_AT=898879893bf1b3361fa783acc5d586d7ff0cfeb5
```

Scope re-derived from `898879893`'s own `ci.yml`, then run **wider** than CI:

| step | result |
|---|---|
| `cargo check -p lix --no-default-features` *(not a CI step)* | **0** |
| `cargo check -p lix --tests` | **0** |
| `cargo check --workspace --all-targets --all-features` *(not a CI step)* | **0** |
| `cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings` | **0** |
| `cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast` | **0** |
| `cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast` | **0** |
| `cargo test -p lix --lib tracked_state::mutation_directory:: -- --test-threads=1` | **0** |

```
EXECUTED_TARGETS test1=25 test2=27
PASSED_test1=3479  PASSED_test2=167  IGNORED_test1=34
SERIAL_module: 6 passed; 0 failed; 2 ignored
```

### Before/after pair, because this PR installs a `#[cfg(test)] #[global_allocator]`

The allocator is a pass-through to `System` — which this crate already used, declaring none otherwise — but it applies to all ~2 690 lix lib tests, so "it perturbs nothing" is a claim that needs evidence rather than an argument. Base run at `898879893` in a **dedicated target dir**, same host, same feature set:

| | executables | passed | ignored |
|---|---|---|---|
| base `898879893` | 25 | 3 478 | 32 |
| this branch | 25 | 3 479 | 34 |

**+1 passed** — the one non-ignored test this branch adds — **+2 ignored**, the two measurement probes, and **no new executable**, since it adds no `[[test]]` target. Nothing else moved.

Arm identity asserted from the logs by test name, not by directory:

```
dirnode_point_read_cost_curve                                    base=0  branch=1
dirnode_point_lookup_v1_vs_v2                                    base=0  branch=1
dirnode_v2_summary_matches_walked_summary_and_rejects_touched_disorder
                                                                 base=0  branch=1
```

*(A first attempt at this witness grepped the base log for `dirnode` and got 38 hits, which read as contamination. All 38 were the worktree path — `dirnode-base-target` ×25, `dirnode-base` ×13. The fix is above: match on test names, which a path cannot contain.)*

### A red gate on the first commit, fixed rather than pushed past

`31ca34401` failed `clippy -D warnings` with three errors, all in its own test-only code and all invisible to every other step, because only `clippy --all-targets` builds the lib **test** target: two `unused_qualifications` and one `dead_code`. Fixed in `3b7b85e95`.

samuelstroschein:
Merged to `main` after a wider-than-CI gate on the two-PR union (#1465 + #1467). #1468 (`e-tswriter`) was deliberately **excluded**.

**Union gated at `0ded298db`** (= `main`@`9e8497bc4` + #1465 + #1467, both merges conflict-free), **`main` re-gated at `f36b3d845`** from a fresh checkout. Trees are byte-identical (`git diff <union> origin/main` → 0 files). Host `ryzen-9950x-I`.

Scope re-derived from the merge target's own `ci.yml`, then run wider — six steps, all `EXIT=0` on both trees:

```
cargo check -p lix --no-default-features                                  EXIT=0
cargo check -p lix --tests                                                EXIT=0
cargo check --workspace --all-targets --all-features                      EXIT=0   (not a CI step)
cargo clippy --profile test --workspace --all-targets --all-features -D warnings   EXIT=0
cargo test --profile test --workspace --exclude lix_e2e --all-features --no-fail-fast -- --test-threads=1   EXIT=0
cargo test --profile test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace --no-fail-fast   EXIT=0
```

```
test1  25 executables / 3491 passed / 34 ignored / 0 failed
test2  27 executables /  167 passed / 21 ignored / 0 failed
```

**Counts reconciled by name, not by total** (`main`@`898879893` = 3478 → −1 #1464 → +13 #1465 → +1 passed / +2 ignored #1467 = **3491 / 34**, predicted before the run):

- #1465 — all **13** `hot_state::typed_slots::tests::*` present and `... ok`.
- #1467 — `dirnode_v2_summary_matches_walked_summary_and_rejects_touched_disorder ... ok`, plus `dirnode_point_read_cost_curve` and `dirnode_point_lookup_v1_vs_v2` `... ignored, measurement probe, not a gate` (+1 passed / +2 ignored).
- #1468 — its **five** test names appear **zero** times in the executed log. That absence, not the total, is the proof #1468 did not come along: the earlier four-way union read 3496 = 3491 + 5.

**#1468 exclusion asserted mechanically**, since `e-tswriter` was stacked on `e-tslots` and merging the wrong ref would have dragged it in silently:

```
git merge-base --is-ancestor 2bc45aaaf c798cd11b   -> exit 1 (FALSE)   # not on the e-tslots tip merged
git merge-base --is-ancestor 2bc45aaaf 3b7b85e95   -> exit 1 (FALSE)
git merge-base --is-ancestor 2bc45aaaf origin/main -> exit 1 (FALSE)   # after the merges
git merge-base --is-ancestor c798cd11b 2bc45aaaf   -> exit 0 (TRUE)    # confirms the stack direction
```

**No protocol bump.** `REPOSITORY_PROTOCOL_VALUE` reads `tracked-default-branch.v68` on `main` at `f36b3d845`, unchanged. Neither of these two PRs requires a bump; only #1468 does, and it is not merged.

Union diff vs `main` is purely additive — 3 files, 1867 insertions, 0 deletions.

PR #1468, opened 2026-08-13
typed-slot writer: a byte-valued payload channel through the head value

samuelstroschein:
Stacked on **#1465** (`e-tslots`, draft, not merged). #1465 landed the typed-slot record format, its decode path and the byte-identity property; it deliberately shipped **no writer**, because the payload channel is `str`-typed end to end and a decode branch keyed on the existing `&str` channel could never be taken. This PR opens the byte-valued channel and writes through it.

## 1. The byte-valued payload channel

Head value **v9 → v10**, adding one header byte at offset 59:

```text
 58   columnar base-coordinate presence (0 or 1)
 59   payload encoding: bit 0 snapshot is a typed-slot record, bit 1 metadata is
 60.. snapshot payload, metadata payload, ...
```

A row is JSON text **or** typed bytes, and which one is **stated in the header, never inferred from the payload**. `TypedSlotsRef::looks_like_record` (a content sniff, from #1465) exists but is not what the storage path keys on: a format that guesses its own discriminant has no way to fail loudly when the guess is wrong.

`HeadSlotView::Typed(&[u8])` / `HeadSlotEncode::Typed(&[u8])` are new variants alongside `Inline(&str)` rather than a reinterpretation of it. Fixed-width numeric payloads are not valid UTF-8, so this is not optional.

The flags byte was already full (deleted, 2×2 slot kinds, untracked, 2 working-diff bits), which is why the discriminant is a new byte rather than a fifth slot kind.

## 2. The writer, and the route to a read

`append_head_value_with_typed_layout` encodes the snapshot as a typed record when the caller resolves a layout for the row's schema. It is wired into the two staging lanes that construct a head value from a live delta — `stage_current_state_with_working_diff_inner` (`hot.rs`, the lane an ordinary `INSERT`/`UPDATE` reaches at commit) and the deferred fresh-hot page builder — both of which have `delta.schema_key` in scope. Every other head-write site keeps its existing signature and writes JSON text.

Two gates on the writer, both mechanisms rather than readings of the schema files:

* **Layouts come from the compiled-in built-in schemas** (`schema/builtin/*.json`, already `include_str!`-ed), restricted to schemas that are closed (`additionalProperties: false`) and whose every declared property is scalar. A layout is therefore a deterministic function of the binary, not of repository content or of the order in which SQL surfaces happened to be built. Of the 14 built-in schemas, the all-scalar ones get typed records; `lix_key_value` does not, because `value` is arbitrary JSON by design.
* **The writer refuses a snapshot carrying a key the layout does not declare.** `encode_typed_slots` writes one slot per *declared* column, so an undeclared key would be silently dropped. Closed schemas make that unreachable; the check is what makes it unreachable *by mechanism*. Declining is always safe — the row is stored as JSON text exactly as today.

### The record now names its own layout

The record header carries an 8-byte **layout fingerprint** (blake3 over the declared names and types, in layout order). Two consequences:

* A read site needs **no schema key at all**: `builtin_layout_for_fingerprint` resolves the record's own layout, so nothing has to be threaded down to materialization.
* `to_json_value` **refuses** a layout whose fingerprint does not match the record's. Without it, a lookup that returned a different column set would decode every slot at the wrong name, silently. `reconstruction_refuses_a_layout_the_record_was_not_written_against` pins this with a same-arity, same-types, different-names impostor — the case a length check cannot catch.

## 3. Reconstruction on demand

`materialize_live_slot` reconstructs canonical JSON from a typed record, so `snapshot_content` still hands every one of its **817 references across 40+ files** the text they expect and **nothing else in this PR has to change**. That is only sound because of #1465's result: reconstruction is byte-identical to the normalized text the record replaced. Content addressing hashes that text, so this is a correctness precondition, not a performance nicety, and this PR keeps a test that fails if it diverges by one byte.

The working-diff fingerprint follows from the same property. `working_diff_slot_fingerprint` hashes the reconstruction of a typed slot, so a row whose before-image is stored as text and whose current version is typed still classifies **Equal** on equal content. An unresolvable record answers `unresolved` rather than guessing — an accelerator may decline, but must never be confidently wrong.

## Evidence

Deterministic, host-independent, no timings.

`a_head_row_for_an_all_scalar_schema_is_stored_typed_and_read_back_byte_identically` stages a `lix_account` row through the writer and reads it back:

| | assertion |
|---|---|
| writer engaged | `TYPED_SLOT_RECORDS_WRITTEN` moved |
| stored form | header byte 59 has the typed bit; the payload does not start with `{` |
| decoded form | `HeadSlotView::Typed(_)` |
| **read served from the record** | `TYPED_SLOT_READS_SERVED` moved |
| **text handed to consumers** | byte-identical to the JSON that was written |

The counters are the point: a decode branch that exists but is never taken looks exactly like a live one in a diff. The **positive control runs in the same test** — `lix_key_value` has a JSON-declared column, so it has no layout, takes the JSON route, and its stored row decodes back to `HeadSlotView::Inline`.

`a_typed_slot_fingerprints_identically_to_the_json_text_it_replaced` pins the dirty/clean agreement described above.

## What is not here

**No performance claim, and no measurement.** In this staged landing the text is reconstructed at materialization, so a typed read pays a reconstruction the JSON read does not; the residue another experiment put at 255–320 ns/row (~48–53% of parse cost, on selective scans) is only available once the *filter* path reads slots directly and the text consumers stop asking for text. That is the next step, not this one. **Anything said here about that residue is inferred-not-run.**

Also untouched: `parse_snapshot_filter_columns`, the entity provider's filter path, `storage_codec.rs`, `tracked_state/*` (owned elsewhere this round).

## Protocol

**This change requires a `REPOSITORY_PROTOCOL_VALUE` bump.** Unlike #1465, which was inert on stored bytes, a writer means head values are genuinely v10 and carry typed payloads — an existing repository must be rejected, not degraded. **The value is deliberately left at `tracked-default-branch.v68` here**, because this round applies one bump once at integration; four separate bumps would be four mutually incompatible repositories.

Note that a change to a built-in schema now also changes a stored layout fingerprint. That is a stored-format change and is carried by the same bump.

## Gate

Scope re-derived from the base sha's own `ci.yml`, then run wider: `cargo check -p lix --no-default-features`, `cargo check -p lix --tests`, `cargo check --workspace --all-targets --all-features` and a `--test-threads=1` test1 lane are none of them CI steps at that sha. Gate evidence, including the new tests asserted by emitted name, is posted as a comment on this PR.

### DIFF
--- a/packages/lix/src/hot_state/tracked_head.rs
+++ b/packages/lix/src/hot_state/tracked_head.rs
@@ -1321,17 +1321,26 @@ fn working_diff_error(message: &str) -> LixError {
 /// 50..54  snapshot payload byte length (big endian u32)
 /// 54..58  metadata payload byte length (big endian u32)
 /// 58      columnar base-coordinate presence (0 or 1)
-/// 59..    snapshot payload, metadata payload, then an optional fixed
+/// 59      payload encoding: bit 0 snapshot is a typed-slot record, bit 1
+///          metadata is. Zero means both payloads are JSON text.
+/// 60..    snapshot payload, metadata payload, then an optional fixed
 ///          checkpoint before-image, then an optional 24-byte base coordinate
 /// ```
 ///
 /// Slot payloads are either inline UTF-8 JSON, a fixed 32-byte `JsonRef`, or
 /// that same fingerprint followed by inline JSON for dirty working-diff rows.
+///
+/// An inline payload is additionally either JSON text or a *typed-slot record*
+/// (`hot_state::typed_slots`), whose fixed-width number payloads are not valid
+/// UTF-8 and so cannot travel through the `&str`-typed inline slot. Which one a
+/// row carries is stated by the payload-encoding byte and is never inferred
+/// from the payload's own bytes: a format that guesses its own discriminant has
+/// no way to fail loudly when the guess is wrong.
 /// Persisting fingerprints only while a row is dirty keeps repeated diff
 /// classification independent of payload size without taxing checkpointed
 /// current state.
-const HEAD_VALUE_VERSION: u8 = 9;
-const HEAD_VALUE_HEADER_BYTES: usize = 59;
+const HEAD_VALUE_VERSION: u8 = 10;
+const HEAD_VALUE_HEADER_BYTES: usize = 60;
 const COLUMNAR_BASE_COORDINATE_BYTES: usize = 16 + 4 + 4;
 const HEAD_VALUE_DELETED: u8 = 0b0000_0001;
 const HEAD_VALUE_SNAPSHOT_SHIFT: u8 = 1;
@@ -1344,6 +1353,8 @@ const HEAD_SLOT_NONE: u8 = 0;
 const HEAD_SLOT_REF: u8 = 1;
 const HEAD_SLOT_INLINE: u8 = 2;
 const HEAD_SLOT_INLINE_FINGERPRINTED: u8 = 3;
+const HEAD_PAYLOAD_TYPED_SNAPSHOT: u8 = 0b0000_0001;
+const HEAD_PAYLOAD_TYPED_METADATA: u8 = 0b0000_0010;
 const HEAD_WORKING_DIFF_DISABLED: u8 = 0;
 const HEAD_WORKING_DIFF_CLEAN: u8 = 1;
 const HEAD_WORKING_DIFF_BEFORE_ABSENT: u8 = 2;
@@ -1356,7 +1367,13 @@ enum HeadSlotView<'a> {
     None,
     Ref(JsonRef),
     Inline(&'a str),
-    InlineFingerprinted { json_ref: JsonRef, json: &'a str },
+    InlineFingerprinted {
+        json_ref: JsonRef,
+        json: &'a str,
+    },
+    /// A typed-slot record. Not UTF-8, and deliberately a separate variant
+    /// rather than a reinterpretation of `Inline`.
+    Typed(&'a [u8]),
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -1558,7 +1575,65 @@ fn working_diff_slot_fingerprint(slot: HeadSlotView<'_>) -> WorkingDiffSlotFinge
             kind: WORKING_DIFF_SLOT_INLINE,
             hash: *json_ref.as_hash_array(),
         },
+        // The fingerprint must agree across the two representations, because a
+        // row's before-image may be stored as JSON text while its current
+        // version is a typed record. It does, and only because reconstruction
+        // is byte-identical to the text the record replaced: hashing the
+        // reconstruction is hashing the same bytes the JSON slot would have
+        // hashed. An unresolvable record answers `unresolved` rather than
+        // guessing -- an accelerator may decline, but must never be confidently
+        // wrong.
+        HeadSlotView::Typed(record) => match typed_slot_canonical_json(record) {
+            Ok(json) => WorkingDiffSlotFingerprint {
+                kind: WORKING_DIFF_SLOT_INLINE,
+                hash: *JsonRef::for_content(json.as_bytes()).as_hash_array(),
+            },
+            Err(_) => WorkingDiffSlotFingerprint::unresolved(),
+        },
     }
+}
+
+/// Rebuilds the JSON text a typed-slot record was encoded from.
+///
+/// The layout comes from the record's own fingerprint, so no schema key has to
+/// be threaded to a read site, and a record whose layout this binary does not
+/// know is an error rather than a positional misread.
+fn typed_slot_canonical_json(record: &[u8]) -> Result<String, LixError> {
+    let parsed = crate::hot_state::typed_slots::TypedSlotsRef::parse(record)
+        .map_err(|error| head_value_error(&format!("typed snapshot slot: {error}")))?;
+    let layout =
+        crate::hot_state::typed_slots::builtin_layout_for_fingerprint(parsed.layout_fingerprint())
+            .ok_or_else(|| {
+            head_value_error("typed snapshot slot names a column layout this build does not know")
+        })?;
+    parsed
+        .to_canonical_json(layout)
+        .map_err(|error| head_value_error(&format!("typed snapshot slot: {error}")))
+}
+
+/// Encodes one snapshot as a typed-slot record, or declines.
+///
+/// Declining is always safe: the row is then stored as JSON text exactly as it
+/// is today. The key check is the load-bearing one -- `encode_typed_slots`
+/// writes a slot per *declared* column, so a key the layout does not declare
+/// would be silently dropped. Closed schemas make that unreachable, and this
+/// check is what makes it a mechanism rather than a reading of the schema
+/// files.
+fn typed_snapshot_record(
+    layout: &crate::hot_state::typed_slots::TypedSlotLayout,
+    json: &str,
+) -> Option<Vec<u8>> {
+    let serde_json::Value::Object(map) = serde_json::from_str::<serde_json::Value>(json).ok()?
+    else {
+        return None;
+    };
+    if map.keys().any(|key| layout.index_of(key).is_none()) {
+        return None;
+    }
+    let record = crate::hot_state::typed_slots::encode_typed_slots(layout, &map).ok()?;
+    #[cfg(test)]
+    crate::hot_state::typed_slots::record_typed_slot_record_written();
+    Some(record)
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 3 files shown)


samuelstroschein:
## Gate — `ryzen-9950x-I`, sha `2bc45aaaf67bd21695a9abd26ab0a8a69e6bc493`

Scope re-derived from the base sha's own `ci.yml`, then run wider. Private worktree `~/claude5/tmp/ts-writer`, run nonce `3120021-1786648812`, logs namespaced by nonce. Counts predicted before the run and reconciled after.

```
RUN_NONCE=3120021-1786648812
HEAD_BEFORE=2bc45aaaf67bd21695a9abd26ab0a8a69e6bc493
DIRTY=0
CI_SCOPE_CONFIRMED_AT=2bc45aaaf67bd21695a9abd26ab0a8a69e6bc493
EXIT_nodefault=0     cargo check -p lix --no-default-features        (not a CI step)
EXIT_libtests=0      cargo check -p lix --tests                      (not a CI step)
EXIT_wsall=0         cargo check --workspace --all-targets --all-features
EXIT_clippy=0        cargo clippy --profile test --workspace --all-targets --all-features -- -D warnings
EXIT_test1=0         cargo test --workspace --exclude lix_e2e --all-features -- --test-threads=1
EXIT_test2=0         cargo test -p lix_e2e --features sdk-tests,plugin-tests,server-protocol,storage-benches,slatedb,root-replay-trace
TEST1_EXECUTABLES=25
TEST1_PASSED=3496
TEST2_EXECUTABLES=27
TEST2_PASSED=167
HEAD_AFTER=2bc45aaaf67bd21695a9abd26ab0a8a69e6bc493
```

### Predicted before running, reconciled after

| metric | base `c798cd11b` | predicted | measured |
|---|---|---|---|
| test1 executables | 25 | 25 (no new `[[test]]` target) | **25** |
| test1 passed | 3491 | 3496 (+5, none `#[ignore]`d) | **3496** |
| test2 executables / passed | 27 / 167 | unchanged | **27 / 167** |

### `NEW_TEST_PRESENT=5`, asserted by emitted name in the executed log

```
hot_state::typed_slots::registry_tests::a_builtin_all_scalar_schema_has_a_layout_and_a_json_declared_one_does_not ... ok
hot_state::typed_slots::registry_tests::a_record_resolves_its_own_layout_from_its_fingerprint ... ok
hot_state::typed_slots::registry_tests::reconstruction_refuses_a_layout_the_record_was_not_written_against ... ok
hot_state::tracked_head::tests::a_head_row_for_an_all_scalar_schema_is_stored_typed_and_read_back_byte_identically ... ok
hot_state::tracked_head::tests::a_typed_slot_fingerprints_identically_to_the_json_text_it_replaced ... ok
```

Provenance rests on the sha assertion inside the gate script, the run nonce printed at both ends, and `HEAD_BEFORE == HEAD_AFTER`. #1465's `FOREIGN_SHA_IN_LOG` check is deliberately not copied: it greps for its own worktree path, so it matches itself and is uninformative.

`--no-default-features` matters here beyond habit: the writer's engagement counters are `cfg(test)` statics, and an ungated reference to one could not compile in that configuration.

samuelstroschein:
## The typed-slot writer does not fire on the staging lane: an ordinary public-API write stores a JSON record, not a typed one

This PR's author stated, correctly and unprompted, that the staging-lane wiring was established **by reading and compilation, not by measurement**. I measured it. The wiring does not produce a typed record in an ordinary repository lifecycle.

**Established by measurement.** Host `ryzen-9950x-I`, throwaway worktree at `2bc45aaaf67bd21695a9abd26ab0a8a69e6bc493`, `HEAD_BEFORE == HEAD_AFTER`, run nonce `3203678-1786650458`.

A probe opened `open_lix().with_storage(Memory::default())`, wrote through the ordinary public path, and then read the **stored bytes** out of `hot_state::ROW_SPACE` — deliberately not `TYPED_SLOT_RECORDS_WRITTEN`, since a counter can be moved by a unit test that never reaches the async storage writer.

### Four phases, zero typed records

```
PHASE=after_ordinary_writes        hot_rows=45 typed_rows=0  working_diff[disabled=1,clean=41,before_absent=3]
PHASE=after_checkpoint             hot_rows=46 typed_rows=0  working_diff[disabled=1,clean=42,before_absent=3]
PHASE=after_post_checkpoint_write  hot_rows=48 typed_rows=0  working_diff[disabled=1,clean=42,before_absent=5]
PHASE=after_reopen                 hot_rows=48 typed_rows=0  working_diff[disabled=1,clean=42,before_absent=5]
```

Two positive arms, both **layout-bearing built-in all-scalar closed schemas** — `lix_account` (via `ensure_account`) and `lix_branch_descriptor` (via the fully public `create_branch`) — plus `lix_key_value` as a negative control. Every witness row was identical:

```
witness=account   snapshot_kind=3 working_diff=2 payload_encoding=0b00000000 typed=false json_text_present=true
witness=branch    snapshot_kind=3 working_diff=2 payload_encoding=0b00000000 typed=false json_text_present=true
witness=key_value snapshot_kind=3 working_diff=2 payload_encoding=0b00000000 typed=false json_text_present=true
```

Values read back correctly (`Text("e90-account-precheckpoint")`), and every stored row carried version byte **10**.

**Scoring this run honestly: it does not discriminate.** Both arms came out non-typed, so the negative control cannot be credited — a control only earns its keep when the positive arm fires. What the run establishes is the absence, not the discrimination.

### Three lanes reach the head; none can emit a typed record

1. **The `BeforeAbsent` discard — `tracked_head.rs:1728-1735`.** Every *fresh* user write is staged with `WorkingDiffBaseline::BeforeAbsent`, because a newly-written row is by construction absent at the checkpoint. `append_head_value_with_typed_layout` is handed the correct layout by its call site and then discards it under exactly that condition (`!fingerprint_inline`). The measured `working_diff=2` on every user-written witness row is this branch being taken.

2. **The seed lanes never call the typed writer at all.** 41 of the 45 rows present after ordinary writes are repository-init rows staged `Clean` — `fingerprint_inline == false`, so the guard in (1) would *permit* a typed record — and they are still non-typed. These are the four lanes named in the module's own comment at `tracked_head.rs:459-474`: `init::plan`'s seed rows, `current_state_delta_from_engine_row`, `functions::state::stage_sequence`, `hot::stage_hot_bootstrap`. Several of those schemas (`lix_account`, `lix_branch_descriptor`, `lix_commit`, `lix_commit_edge`, `lix_checkpoint`, `lix_branch_ref`) do have layouts.

3. **The restage lane takes no layout parameter.** A checkpoint does re-encode rows — the `lix_key_value` witness moved from `kind=3, working_diff=2` to `kind=2, working_diff=1` — but that path is `reencode_head_value_with_baseline` → `encode_head_value_parts`, whose signature has no layout argument at all.

### Why this blocks the merge

Version byte **10 is already live on disk**, and `decode_head_value` hard-errors on any other version (`"unsupported row format version {n}"`); the fixed header also grew 59 → 60 bytes, shifting every payload offset. So this PR imposes the full cost of a stored-format break — an existing repository must be *rejected*, which is what the pending `REPOSITORY_PROTOCOL_VALUE` bump would carry — while delivering none of the benefit, because the payload channel it opens is never written to.

The second consequence the author recorded (a change to a built-in schema now also changes a stored layout fingerprint) is currently **latent** for the same reason: no fingerprint is ever stored.

**This must not merge as-is.**

### Why the suite looks like it already covers this

This test passes on the branch:

```
hot_state::tracked_head::tests::a_head_row_for_an_all_scalar_schema_is_stored_typed_and_read_back_byte_identically ... ok
```

Its name asserts precisely what the probe above disproves at the storage layer. It is green because it calls `append_head_value_with_typed_layout` **directly, with a non-dirty baseline** — so it exercises the *encoder*, not the *lane*. Every real staging path arrives with `BeforeAbsent` and takes the discard branch in (1), which this test never reaches.

Worth fixing alongside the writer: a test whose name claims stored-format behaviour should fail when no record is stored. A disk-level assertion on `ROW_SPACE` — as used here — is the shape that would have caught this.

### Not attempted here

No fix. It spans three staging lanes, and the `BeforeAbsent` guard exists for working-diff fingerprinting rather than by oversight — so the scope question (does the typed lane need its own fingerprinted-inline framing, or does it stay excluded from dirty rows?) is a design call, not a patch.

### Union gate, for the record

`e-tswriter` is linearly stacked on `e-tslots`, so their "union" is a no-op. A throwaway union of `2bc45aaaf` + current `origin/main` (`9e8497bc4`) + #1467 merged with zero conflicts and gated green through `cargo check -p lix --no-default-features`, `cargo check -p lix --tests`, `cargo check --workspace --all-targets --all-features`, and `clippy --all-targets --all-features -D warnings`. Test lanes also green: **25 executables / 3496 passed / 34 ignored / 0 failed**, test2 27 / 167.

That total is *not* self-evidencing — 3496 is simultaneously the union's prediction and `e-tswriter`'s standalone figure, because #1464's −1 cancels #1467's +1. The terms were verified by test name instead. **So the stack is otherwise sound; the defect above is the only thing blocking it.**

*No timings are quoted from this branch: in this staged shape a typed read pays a reconstruction cost the JSON read does not.*

PR #1470, opened 2026-08-13
feat(lix-schema): canonical value encoding for the typed-row body

samuelstroschein:
**Stacked on #1469 (`agent/postgres-derived-schema-v1`), pinned at `7d587bde020af2a2e5733d7009645bdadb2bb67b`.** That branch is being rebased onto `main` by another agent; this branch deliberately does not touch its refs and will need retargeting once the rebase lands. Draft, not for merge.

Run nonce `vlayout-8813`. Host `ryzen-9950x-IV`.

## What this adds

`lix_schema::value_layout` — the **value half** of the typed-row layout, and the tests that make its load-bearing property checkable. The metadata half was settled by a previous job and is unchanged here.

The property under test: **one logical value produces exactly one byte string.** Two byte strings for one value means two content addresses for one snapshot, which breaks dedup and identity.

| type | body image | canonicalisation | rejected |
|---|---|---|---|
| `boolean` | 1 fixed byte | — | any byte but `0x00`/`0x01`, on decode |
| `bigint` | 8 fixed bytes, BE two's complement | — | — |
| `double precision` | 8 fixed bytes, BE IEEE-754 | `-0.0` → `+0.0` | NaN (any payload), ±Inf |
| `uuid` | 16 fixed bytes, RFC 4122 order | textual spelling parsed away | unparseable text |
| `text` | raw UTF-8, var area | **none, deliberately** | interior NUL |
| `jsonb` | canonical semantic JSON | keys sorted, integral f64 → i64, spelling normalised | NUL, non-finite number |

Two decisions worth naming:

- **`-0.0` is the one real canonicality break in the value set.** `-0.0 == 0.0` under IEEE-754 and under Rust's `PartialEq`, so they are one value — but their bit patterns differ, and `-0.0` is `is_finite()`, so it passes the existing row validation untouched. The encoder folds the sign of zero and the decoder refuses a `-0.0` slot.
- **`text` is not Unicode-normalised, and that is correct.** NFC "é" and NFD "é" are distinct PostgreSQL `text` values; normalising would make the encoding non-injective with respect to what the user stored. `canonicality_text_is_not_unicode_normalised` pins it, so a future upstream normaliser surfaces here rather than silently in dedup behaviour.

## Evidence

Each canonicality test pairs a **two-construction-path equality** with a **positive control that must differ**, so no test can pass by encoding nothing.

The suite was **mutation-tested** — six deliberate breaks of canonicality, five killed:

| mutant | result |
|---|---|
| drop the `-0.0` fold | killed (2 tests) |
| drop the integral-f64 fold in jsonb | killed |
| accept any non-zero byte as boolean `true` | killed |
| allow non-finite doubles through | killed |
| stop clearing the output buffer (stale bytes in NULL slots) | killed (2 tests) |
| **drop the jsonb key sort** | **SURVIVED** |

The survivor is a real finding, not a test gap, and it is written into the suite as `canonicality_jsonb_key_order_mechanism_is_pinned`: `serde_json::Map` without the `preserve_order` feature **is** a `BTreeMap`, so it already serialises in sorted key order and the explicit `sort_by` is currently redundant. Key-order canonicality therefore rests on the *absence of a cargo feature* anywhere in the workspace — and cargo unifies features across the whole graph. The added test asserts raw `serde_json` still sorts, so enabling `preserve_order` fails loudly instead of quietly shifting which mechanism is load-bearing.

## Sizes, values included

24 fixture schemas, `packages/lix-schema/fixtures/current/*.json`, printed by `tests/row_sizes.rs`. Values are **synthesised** (declared `examples[0]` where present — only 3 fixtures carry any — otherwise a type-appropriate realistic value), so this prices the *layouts* against a fixed population, not the population.

- **Total 3157 B → 1406 B, 2.25×.** Mean 131.5 → 58.6 B/row.
- Metadata only: **40.50 → 4.75 B/row, 8.53×.**
- Range: 1.56× (`lix_change`, `csv_row` — value-dominated) to 48× (`lix_commit`, one PK-only column, which the body elides entirely to a bare header byte).

## Gate

```
HEAD_BEFORE = HEAD_AFTER = 7d587bde020af2a2e5733d7009645bdadb2bb67b (pre-commit)
cargo test -p lix-schema --all-targets      EXIT=0   16 + 2 + 4 + 1 passed
cargo clippy -p lix-schema --all-targets -- -D warnings   EXIT=0
```

Scope is `packages/lix-schema` only; the crate has no dependents that this touches (a new `pub mod`, no existing item changed).

## Not done here

- `TAG_ABSENT` removal — see the accompanying report. Schema v1 makes "absent" unrepresentable, and the sites are confined to `packages/lix/src/hot_state/typed_slots.rs` plus one e2e example, but the removal belongs in `packages/lix` and would collide with in-flight work.
- Encode/decode timing. Deterministic byte counts only; no rate is claimed.


samuelstroschein:
Consolidated into #1469. The canonical typed-row codec commit was ported, extended for int8/float8/timestamptz, tested, and published at 2cf539744e7864f79bf1994e002f47cfd3281dc0.

PR #1472, opened 2026-08-13
build(rust): isolate the embeddable SDK from artifact dependencies

samuelstroschein:
## Summary

- split the stable embeddable Rust SDK/storage workspace from CLI/E2E Cargo artifact builds
- keep CLI/E2E and their Wasm `bindeps` in `tooling/Cargo.toml`
- run both workspaces in CI and add a clean Cargo 1.93 local-path consumer gate

## Reproduction and profile

Before this change:

```text
cargo +1.93 metadata --manifest-path packages/lix/Cargo.toml --no-deps
error: `artifact = …` requires `-Z bindeps` (plugin_csv)
wall=0.02s, exit=101
```

After this change:

```text
wall=0.01s, exit=0, max RSS=30,656 KiB
```

No Rust production source or runtime path changes in this layer, so runtime performance is byte-identical. Root and tooling lockfiles preserve the prior resolved versions for their respective graphs.

## Gates

- Cargo 1.93 SDK metadata: PASS
- nightly tooling metadata with `bindeps`: PASS
- `cargo check -p lix --lib --no-default-features`: PASS (2m46s cold)
- publish-surface validator: PASS
- release script tests: 10/10 PASS
- YAML parse and diff-check: PASS

This is stack 1/3 for the Rust embedding P0s.

PR #1474, opened 2026-08-14
fix(rust): make core SDK scheduling runtime-neutral

samuelstroschein:
## Summary

- remove `tokio::runtime::Handle::current()` from the transaction commit coordinator
- run engine-owned commit, checkpoint-GC, and invalidation workers on Lix-owned executor-neutral threads
- make core `open_lix`/execute/close work under plain `futures_lite::future::block_on`
- add an external Filesystem `open_lix` + `start_sync` plain-`block_on` contract

This is stacked on #1472 so the external SDK workspace is stable-Cargo addressable first.

## Runtime contract

Core production scheduling no longer requires an embedding Tokio runtime. The optional server-protocol transport remains Tokio-specific; it is feature-scoped transport code rather than the core embedded SDK.

The temporary 16 MiB engine-owned worker stack is deliberately retained in this PR. Removing that workaround is the next stacked default-stack/future-flattening cut, so runtime neutrality does not silently change stack behavior.

## Gates

- `cargo check -p lix --lib --no-default-features`: PASS
- `cargo check -p lix --tests --no-default-features`: PASS (inherited warnings only)
- `cargo test -p lix --test runtime_contract --no-default-features public_sdk_opens_writes_and_reads_without_a_tokio_runtime`: PASS, 1/1, 0.04s
- `git diff --check`: PASS
- owned runtime residue: no production `Handle::current`; core background sites no longer call `tokio::spawn`
- repository-wide fmt check remains RED only on pre-existing files outside this diff
- Filesystem runtime test: source/type path added; execution not claimed because two bounded attempts were consumed compiling `librocksdb-sys` C++ and timed out before Rust/link/test

## Performance guard

Byte-identical core public custom-runtime open/write/read/close test, debug binary, 20 fresh processes per side:

| | main baseline | candidate |
|---|---:|---:|
| wall median | 50 ms | 50 ms |
| wall mean | 47.0 ms | 50.5 ms |
| median max RSS | 64.65 MiB | 65.12 MiB |
| max RSS | 65.08 MiB | 65.45 MiB |

The median is unchanged; RSS delta is ~0.7%. The 3.5 ms mean difference is within the coarse process-level/noisy debug sample and is not material. Raw TSV SHA-256: baseline `12f2b36b66c01569a8893d49856f898aaaa11a36cd03f062c4176e11254`, candidate `5a1cc5b23069be783f343dd74f2f584788ec0230d89da6bd3599c3633667d13d`.

## Identity

- parent: `f886efdaa9aba5bca14b3f3dc7af049f2d62cb0f`
- head: `8d78766a2c12633e9ad1a1d14011211f34a31faa`
- tree: `b9d0f217f44009b152196670a8df5001e7409d00`
- diff SHA-256: `14ce90ac0ccf5e96cacb4cf3ef3aee543cd0fbe0a535dc458cbdaadb7fe1ad72`
- format-patch SHA-256: `e4d3298e3678aa5d39f5425beef522ae1ada68eda8f58a75710db8379932abc1`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2252,6 +2252,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e5c1b78ca4aae1ac06c48a526a655760685149f0d465d21f37abfe57ce075c6"
 
 [[package]]
+name = "futures-lite"
+version = "2.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f78e10609fe0e0b3f4157ffab1876319b5b0db102a2c60dc4626306dc46b44ad"
+dependencies = [
+ "fastrand",
+ "futures-core",
+ "futures-io",
+ "parking",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "futures-macro"
 version = "0.3.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3040,6 +3053,7 @@ dependencies = [
  "fastcdc",
  "flate2",
  "futures-core",
+ "futures-lite",
  "futures-util",
  "getrandom 0.3.4",
  "globset",
@@ -3092,6 +3106,7 @@ dependencies = [
 name = "lix-storage-filesystem"
 version = "0.11.0"
 dependencies = [
+ "futures-lite",
  "lix",
  "lix-storage-rocksdb",
  "notify-debouncer-full",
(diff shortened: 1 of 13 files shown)


samuelstroschein:
Clippy qualification: `cargo +nightly-2026-05-21 clippy --offline -p lix --lib --tests --all-features -- -D warnings` PASS (exit 0, 3m23s).

PR #1558, opened 2026-08-18
Re-handshake once on SESSION_GONE and SERVER_CLOSED

samuelstroschein:
Protocol sessions are an in-process HashMap. 410 `LIX_ERROR_PROTOCOL_SESSION_GONE` is a miss (unknown / idle-expired / closed id). 503 `LIX_ERROR_PROTOCOL_SERVER_CLOSED` is the same server going away (`close()` / host `recover()`).

The JS remote client now opens a new session by default:

1. `GET /lix/v1/` without `Lix-Session-Id`
2. Restore the pinned branch via `activeBranchId`
3. Restart observations
4. Retry the failed RPC once

Transaction-scoped calls are not retried. Persistent failure after reopen still propagates. Capacity 503 is not treated as session death.

This is the session-death fix. Hosts must not persist the HashMap or remount UI on 410.

### DIFF
--- a/packages/js-sdk/src/remote/client.test.ts
+++ b/packages/js-sdk/src/remote/client.test.ts
@@ -1313,6 +1313,51 @@ test("a gone protocol session mutation opens a new handshake and retries", async
 	await lix.close();
 });
 
+test("a closed protocol server opens a new handshake and retries", async () => {
+	let handshakeCalls = 0;
+	let executeCalls = 0;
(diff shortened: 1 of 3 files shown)


samuelstroschein:
Closing as superseded.

The JS recover-once path landed in #1557. SERVER_CLOSED recovery belongs in the one Rust protocol client (#1560), not a second JS client on the same files.

PR #1559, opened 2026-08-19
Prototype repository-owned Cedar file permissions

samuelstroschein:
## Prototype goal

Prove whether repository-owned Cedar policies make file sharing maintainable while leaving the policy model extensible to each company.

## What this proves

- Canonical policy files live under `/.lix/permissions/` and are normal versioned Lix files.
- Standard sharing is stored in the built-in `lix_permission_grant` resource model.
- Grant rows are always tracked and global: their authoritative state lives on the global branch and is visible from every working branch.
- Grants cover repository, directory, file, table, and row resources, using stable IDs rather than paths.
- Default access levels (`viewer`, `commenter`, `contributor`, `editor`, `manager`) are translated into Cedar permits.
- Repository-authored Cedar remains authoritative and extensible; a custom `forbid` can override a standard grant.
- Exact structured file reads authorize the active account against the stable `lix_file.id`, so renames preserve grants.
- Cedar default-deny keeps unlisted files private.
- Existing repositories register newly shipped built-in schemas on open, in both the global catalog and every existing branch-local catalog.
- Reserved `lix_*` registrations remain protected: only the system account may install a definition, and only when it exactly matches a shipped built-in schema.
- A bundled WASI plugin validates Cedar source files and exposes transparent, read-only `cedar_permission_source` rows.

## Resource identity

- repository: the repository ID
- directory: stable `lix_directory_descriptor.id`
- file: stable `lix_file.id`
- table: canonical `[schema_key, file_id]`
- row: canonical `[schema_key, file_id, row_pk]`

Directory and file grants inherit to descendants. File grants cover the tables and rows owned by that file; table grants cover their rows.

## Maintainability findings

- Directly linking the full `cedar-policy` crate is not safe for Lix today: its validator enables `serde_json/preserve_order`, which changes Lix canonical JSON through Cargo feature unification.
- The engine links `cedar-policy-core = 4.2.2`; the full validator stays isolated in the WASI plugin.
- Browser SDK raw WASM: 54,652,323 B baseline -> 54,684,440 B prototype (+32,117 B, about 31 KiB).
- Cedar validator/projection WASI plugin: 5,321,730 B raw.

## Intentionally deferred

- Same-snapshot authorization + protected read
- Compiled policy/entity cache
- `managePolicies` enforcement and creator bootstrap/recovery
- Arbitrary SQL or row-level filtering
- Whole-policy-set static schema validation
- Share/unshare convenience API and UI

## Verification

- `cargo test -p lix-schema`
- `cargo test -p lix permission_grants --no-fail-fast`
- `cargo test -p lix authorization --no-fail-fast`
- system-session built-in upgrade regression test
- ordinary-session reserved namespace rejection test
- live existing-repository upgrade through Lixray
- browser JS SDK release WASM build and native binding build

### DIFF
--- a/.changenotes/cedar-permission-grants.md
+++ b/.changenotes/cedar-permission-grants.md
@@ -1,0 +1,7 @@
+---
+type: minor
+---
+
+Added repository-owned Cedar permissions with a global `lix_permission_grant` sharing model.
+
+Tracked grants now express account, group, or anonymous access to repositories, directories, files, tables, and rows. The grants are shared across every branch and remain extensible through repository-authored Cedar policies.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,6 +95,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ar_archive_writer"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "73cd58deff2140a0a8eae87e417bd01db68a33e148aa93d1e8cd837e55e312b6"
+dependencies = [
+ "object",
+]
+
+[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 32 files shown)


samuelstroschein:
End-to-end consumer: opral/lixray#240 vendors this exact commit, removes anonymous SQL, and exercises Cedar-authorized exact file reads through Lixray's public transport and Railway preview.

PR #1639, opened 2026-08-26
Hydrate point-in-time reads on sparse sync replicas

samuelstroschein:
`lix_state_at` and `lix_diff` (the #1637 point-in-time machinery) hard-failed on sync replicas instead of hydrating like `lix_history` does. Found while adopting the API in LixRay: opening a checkpoint on a freshly bootstrapped replica died with `commit … has no tracked-state authority` / `filesystem descriptor references missing directory …`.

**Fixes** — both anchor gates in `state_at.rs` / `diff.rs` now classify absence as the typed `LIX_SYNC_HISTORY_REQUIRED` demand (via `missing_commit_state_manifest_error`, plus an explicit deferred-marker check when the manifest arrived hot), so the session hydrates and retries.

**New e2e coverage** (`sync_mode`), all passing:
- fresh replica reads `lix_state_at` on both filesystem relations at a cold checkpoint, resolves paths via root diff, renders the checkpoint-to-checkpoint span (was: hard failure);
- partially hydrated replica (bounded anchor-boundary lookup first);
- full lineage: v71-authored fixture → v72 partial checkpoints (generated from `4816fdba5`) → migrated here → served → sparse replica reads.

**New golden migration coverage** (`migration_v72_filesystem`): a v72 fixture with directories captured by full and partial checkpoints — verified consistent on the authoring engine — migrates to v74 with checkpoint trees intact, and a post-migration partial checkpoint authors a cumulative consistent tree.

**Known remaining gap** (reproduced live, not covered by any generated fixture): a repository whose rows predate the v71 generator still fails the same way against its real server copy even from a fresh fixed-engine replica — the durable row-pk tree served for one checkpoint lacks a directory entry its file rows reference. Turning that repository's server storage into a snapshot fixture (raw slatedb iteration → `Memory`) is the next step; details in the LixRay-side investigation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


samuelstroschein:
Not shipping — superseded by #1638's complete-snapshot model. The branch stays as local investigation tooling (repro tests, debug export, golden migration fixtures), rebased onto dev for reference in case a repair migration is needed.

PR #1645, opened 2026-08-26
Release the v75 complete-snapshot engine to main

samuelstroschein:
Promotes the dev branch to main — 22 commits:

- **Point-in-time state reads** — `lix_state_at` table function (#1637)
- **Complete snapshot commits** — format v75, CommitRecord v6 with `base_commit_id` (#1638)
- **v74→v75 in-place migration** with filesystem-closure repair for v72-era partial-checkpoint corruption (#1640)
- **Review hardening** (#1642): truncation-checked repair scans, `format_version` guard on the idempotent rewrite skip, clock-skew clamp for base inference, tree-wide directory recovery, per-source-version `-commit-rewrite` crash-fence markers, observe base-refresh generation pinning, and deletion of the unreachable pre-v72 migration chain (~8.9k lines)
- **Armed-observation × switch_branch deadlock fix** (#1641) with regression test
- **Observation resilience under sync churn** (#1643, #1644): observe streams re-evaluate on transient coherent-read expiry, and the lazy base refresh retries expired reads under the shared bounded policy — both found by lixray's deployed two-client sync E2E

Validated downstream: lixray dev runs this head end-to-end (opral/lixray#291) with the full deployed preview E2E green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1647, opened 2026-08-26
fix: ancestry-resolved repairs, provenance-correct HOT diffs, switch rollback

samuelstroschein:
Fixes the three findings from the #1645 release review, each verified against the code and covered by a discriminating regression test:

**P1 — repair injected a head's newer directory version into older commits.** The closure repair resolved missing directories first-writer-wins from branch-head trees, so a directory renamed/moved after the corrupted checkpoint would rewrite that commit's historical name and parent. The repair now collects every distinct row version from every commit tree and, per repaired commit, admits only versions whose owning commit is in that commit's ancestry — causal parents plus the composed global base lineage from `base_commit_id` (decodable because the v6 rewrite runs earlier in the same chain) — newest admissible wins. An id with no admissible version fails closed instead of guessing. The golden fixtures exercise the between-checkpoints creation case (`/brand`) and still pass, as do all 20 `sync_mode` e2e tests.

**P2 — HOT working-span diffs lost global provenance.** With an identity-only projection of `from/to_lixcol_global`, the HOT checkpoint-to-head path was selected and returned empty provenance sets — the regression test shows it reported `[NULL, false]` for a local edit shadowing an inherited global row (the global before-side vanished entirely) where the cold path reports `[true, false]`. Provenance projections now take the composite overlay route, mirroring the existing `retain_payloads` gate.

**P2 — switch_branch left the session on the target branch after a failed refresh.** The selector must move before the observer bump (the refresh's staleness gate keys off it), so an error now rolls the selector back and re-notifies observers; the strict regression test injects a write failure into the boundary refresh and asserts the session stays on its previous branch and can retry.

Full `lix --lib` (2887 passed), migration golden tests, `sync_mode` 20/20, CI-config clippy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/lix/src/migration/api.rs
+++ b/packages/lix/src/migration/api.rs
@@ -510,6 +510,122 @@ where
     Ok(members_injected)
 }
 
+fn migration_snapshot_text(
+    row: &crate::tracked_state::MaterializedTrackedStateRow,
+    field: &str,
+) -> Option<String> {
+    let snapshot = row.decoded_snapshot.as_ref()?;
+    match snapshot.row.get(field) {
+        Some(lix_schema::Value::Text(value)) => Some(value.clone()),
+        Some(lix_schema::Value::Uuid(value)) => Some(value.to_string()),
+        _ => None,
+    }
+}
+
+struct CommitGraphNode {
+    parents: Vec<crate::changelog::CommitId>,
+    base: Option<crate::changelog::CommitId>,
+}
+
+/// Every commit whose state the given commit resolves: the causal parent
+/// closure plus, for local commits, the pinned global base lineage their
+/// trees compose beneath the local overlay.
+fn commit_ancestors(
+    commit_id: crate::changelog::CommitId,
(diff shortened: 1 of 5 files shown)


samuelstroschein:
Pushed follow-ups for both findings:

- **Causal ranking:** versions now rank by strict-ancestor supersession over the commit graph (parents + composed base lineage), never by `created_at` — unit-tested with a descendant version carrying an *older* wall clock than its stale ancestor. Causally incomparable merge-ancestor versions **fail closed** with an explicit error rather than consulting a durable conflict winner: the migration writes permanent history, and the same fail-closed rule the release review set for base inference applies — refuse and name the id rather than pick a winner the repaired commit never observed.

- **Switch serialization:** rather than a conditional rollback (which still lets a concurrent switch to the *same* target be reverted by a failing sibling), a session-shared lock in `SessionBranch` now serializes the whole checkout — selector move, boundary refresh, and rollback are atomic with respect to other clones' switches. `switch_branch` is the only runtime selector mutator, and nothing inside the critical section takes the lock again, so no new cycles.

PR #1650, opened 2026-08-27
fix(opfs): preserve sessions across owner handoffs

samuelstroschein:
## Summary

- persist the OPFS storage-session capability in adapter-owned SQLite metadata
- restore the same capability when the package-owned owner worker hands the repository to a successor
- keep tokenless legacy operations fenced after a handoff
- add a two-worker browser regression test covering the actual backend/Web Lock lifecycle

## Why

PR #1649 made storage sessions mandatory so migrations can fence stale handles. A deployed LixRay multi-tab test exposed that the OPFS token was scoped to an owner-worker instance instead of the storage generation. Navigating the owner tab restarted the backend, minted a new token, and incorrectly fenced healthy peer tabs.

## Validation

- OPFS browser suite: 38/38
- OPFS typecheck
- package dry-run validation
- packed Vite production smoke

Follow-up required by opral/lixray#298.

### DIFF
--- a/.changenotes/preserve-opfs-sessions-across-owner-handoffs.md
+++ b/.changenotes/preserve-opfs-sessions-across-owner-handoffs.md
@@ -1,0 +1,7 @@
+---
+type: patch
+---
+
+OPFS repositories remain writable across browser-tab navigation and owner-worker handoffs.
+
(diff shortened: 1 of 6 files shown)


samuelstroschein:
@codex review

samuelstroschein:
@codex review

PR #1651, opened 2026-08-27
refactor(sql): unprefix diff payload columns

samuelstroschein:
## Summary
- hard-cut lix_diff payload columns from lixcol_diff_type/lixcol_row_count to diff_type/row_count
- update every in-repo query, test, benchmark, example, and documentation consumer with no compatibility aliases
- codify the lixcol invariant: tracked-row and history system metadata stays fenced, while relation payload remains ordinary
- add catalog and provider guards against unknown or duplicate lixcol system fields

## Design note
History metadata and lixcol_row_pk intentionally retain the system prefix. Removing it would collide with valid relation payload columns such as depth, row_pk, and is_deleted; the Excalidraw schema already contains is_deleted. Diff payload is collision-safe because source columns are side-qualified as from_*/to_*.

## Validation
- cargo test -p lix --lib --features all-simulations (3348 passed, 26 ignored)
- cargo test -p lix --doc (10 passed)
- cargo test -p lix --lib relation_diff_ (11 passed after rebase)
- cargo test -p lix --lib lixcol_names_are_reserved_for_system_metadata
- cargo clippy -p lix --lib --features all-simulations -- -D warnings
- cargo check --manifest-path packages/e2e/Cargo.toml --all-targets
- cargo fmt --all -- --check
- git diff --check

Three independent sub-agent reviews completed; all findings were addressed and final reviews are clean.

### DIFF
--- a/.changenotes/relation-column-naming-hard-cut.md
+++ b/.changenotes/relation-column-naming-hard-cut.md
@@ -1,0 +1,9 @@
+---
+type: minor
+---
+
+Standardized relation payload column names in `lix_diff()`.
+
+Diff queries now use `diff_type` and `row_count`; `lixcol_diff_type` and
+`lixcol_row_count` were renamed without compatibility aliases. The `lixcol_`
+prefix remains reserved for engine-owned system metadata.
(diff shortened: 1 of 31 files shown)


samuelstroschein:
Addressed the schema-discoverability feedback in 90d43c83c:

- the SQL surfaces docs now state the naming invariant in one explicit sentence
- near-miss errors are covered to retain DataFusion suggestions
- unrelated misses are covered to retain the complete Valid fields are ... listing with diff_type and row_count

PR #1810, opened 2026-09-16
fix(sdk): preserve admission across canceled sync requests

samuelstroschein:
Foreground partial-replica hydration cancels the background descriptor request. The SharedWorker previously cleared verified credentials on that cancellation, forcing another admission round trip before the next native read. Preserve verification for cancellation while retaining invalidation for other transport failures and the existing rotation/rejection paths.

A real Chromium SharedWorker regression (stub engine binding, real worker/RPC/HTTP transport) makes five canceled requests followed by remote reads: admission calls fall from six to one, an 83.3% deterministic reduction. This is not a production-latency claim. Existing same-account rotation, account-drift rejection, denied credentials, and offline-local-work assertions pass.

Validation: Chromium admission regression, 40 focused SDK unit tests, and SDK TypeScript check pass. No storage or wire-format change; existing repositories need no migration.

First draft in the on-demand partial-replica optimization stack. The experiment ledger records acceptance rules, bounded opening and normal awaited-query invariants, and the five-consecutive-non-improvements stopping rule. Related: opral/lixray#416.


samuelstroschein:

PR #1811, opened 2026-09-16
bench(sync): profile public awaited partial history over HTTP

samuelstroschein:
Add a reproducible history benchmark using a fresh RocksDB partial replica, real loopback HTTP, and the canonical authority handler. Applications issue ordinary awaited SQL; hydration remains internal. Reuse exported synthetic snapshots across revisions, verify ordered results against the authority, assert bounded opening without native reads, and verify warm history with the server unavailable.

The paired runner alternates baseline/candidate order, retains failures, and reports seeded bootstrap intervals. Cases cover 64-checkpoint dense/sparse history, depth controls, 16,000 unrelated files, and an unopened 8 MiB blob. An optional fixed RTT tests sensitivity to round trips.

Validation: benchmark builds; dense4 smoke, eight dense64 samples, three sparse64 samples, and three samples per size control pass correctness/opening/offline assertions. Unoptimized dense64 consistently needs 381 native requests; sparse64 needs 213. A ninth dense64 run reached the existing 30-second read deadline under concurrent development load; this failure is retained and timing is not presented as a production measurement. Optimized paired runs follow in the next experiment.

Stacked on #1810. No production engine, protocol, or storage-format change.


samuelstroschein:

PR #1813, opened 2026-09-16
refactor(sync): normalize native metadata demands

samuelstroschein:
Singleton and batch metadata misses currently duplicate residency and hydration control flow. Normalize both diagnostic formats into one metadata demand and reuse the existing bounded batch path, preserving the original error, epoch checks, immutable-byte validation and write-frontier preparation.

Metadata demand variants fall from two to one; residency and hydration branches each fall from two to one. No protocol, storage or application API change. Ten alternating pairs show unchanged request counts and a small cold-history cost: dense 0.78% slower (95% interval 0.29–1.35% slower), sparse 0.57% slower (interval includes zero). Accepted for simpler control flow, with no speedup claim or material opening/warm regression.

Validation: 4,074 all-simulation tests passed (74 skipped), 10 doctests passed; identical ordered authority results, offline warm reads and bounded opening across history/width/blob controls. Reproducible evidence is in docs/partial-sync-experiments/e05. Also retains the rejected E04 experiment and its measurements as documentation only.

Stacked on #1812. Consecutive completed non-improvements reset to zero. Draft; not deployed.

PR #1814, opened 2026-09-16
perf(sync): hydrate bounded first-parent metadata segments

samuelstroschein:
Cold partial-replica history discovers parent metadata one network round at a time. Add a bounded first-parent metadata selector, triggered by typed SQL history demand, that returns up to 16 commits and optional state headers through the existing leased, epoch-fenced installer. Opening remains independent of history size; applications keep normal awaited SQL.

Ten alternating paired runs against #1813: dense64 native requests 320→262 and cold history 19.29% faster (95% interval 18.55–19.67%); sparse64 152→94 and 38.82% faster (38.07–40.00%). Opening/warm controls show no material regression. Sparse LIMIT 1 is 20.74% faster but transfers 29.5 KB instead of 20.7 KB; this bounded overfetch is documented. Wide-tree dependencies and SQL replay remain targets.

Validation: 4,315 all-simulation/server-protocol tests passed (84 skipped), 10 doctests, 45 SDK unit tests, TypeScript and the Chromium admission regression. Tests cover lease/session bounds, path/epoch validation, retained gaps and speculative corruption without masking required-read errors. Real protocol-17 partial RocksDB replicas reopen with protocol 18, preserve ordered results and work offline when warm, without resets.

Breaking protocol change: SDK and server must upgrade together. Storage format81 and partial receipt v2 are unchanged. The existing SDK admission cache requires online re-admission after a protocol upgrade. Full raw synthetic measurements, manifests and controls: docs/partial-sync-experiments/e06.

Stacked on #1813. Accepted for measured performance; consecutive non-improvements remain zero. Draft, not deployed.


samuelstroschein:

PR #1815, opened 2026-09-16
fix(sync): retain descriptor watch across foreground hydration

samuelstroschein:
Foreground hydration canceled and recreated the descriptor long poll for every missing native input. Retain only the pending network future, keyed by full admission and cursor, while keeping reconciliation cancellable. Also clear the forced-refresh flag after successful expired-lease recovery publishes a new admission; otherwise every subsequent native miss refetches the descriptor.

Ten alternating fresh-replica pairs per case reduce history HTTP attempts from 524 to 262 (dense64) and 188 to 94 (sparse64), exactly 50% in every pair. Native requests and response bodies are unchanged. Paired cold-history latency improves 0.94% [0.33%, 1.46%] dense and 0.08% [-0.90%, 1.45%] sparse: no substantial latency gain is claimed. Three prepared-replica recovery pairs reduce fresh descriptor reads from 262/94 to one. LIMIT1 and dense4 controls also halve HTTP attempts.

All 4,315 all-simulation/server-protocol tests and 10 doctests pass. Opening, file selection and offline warm controls exclude a 10% timing regression; wide16k/blob8MiB opening stays at two foreground requests, zero native reads and <=8KiB response bodies. No protocol, SDK, storage or migration changes.

Adds one pending network-future slot; accepted for request reduction and correctness, not architecture simplification. Measurements, source/binary provenance, reproduction and tradeoffs are in docs/partial-sync-experiments/e07/. Raw captures and private HAR data are excluded.

Stacked on #1814. Draft experiment E07; consecutive substantive non-improvements remain zero.


samuelstroschein:

PR #1848, opened 2026-09-18
fix: query partial history inventories through authority

samuelstroschein:
## Summary

- let partial-replica `lix_commit` and `lix_change` standalone reads retry against the configured authority when local completeness is unsupported
- keep mixed/local queries local so pending replica edits are not silently omitted
- keep the authority session lifecycle in Lix, with a wasm-safe per-request client
- add a patch changenote

## Local verification

- `cargo nextest run --config-file .config/nextest.toml --profile ci --cargo-profile test --workspace --all-features --lib --tests`: 5,398 passed, 96 skipped
- `cargo test --doc --profile test --workspace --all-features --no-fail-fast`
- `node scripts/ci-build-sdk.mjs native`: native, migration-native, wasm, migration-wasm, and plugins passed
- changenote, publish-surface, and server-protocol validators passed


samuelstroschein:
Addressed in `cec0bd305`: authority fallback now rejects queries containing session-dependent execution-slot functions (`lix_active_account_id`, active branch/head, and root commit), including nested expressions. Added focused classifier tests; the full current-main Rust nextest scope passes locally (5404 passed, 96 skipped).

